PR #3, opened 2026-05-06
Reorganize the command tree and introduce config

### LINKED ISSUE #7309: Proposal: Stop using prefixes to differentiate CLI command maturity

Currently, the Crossplane CLI (crank) has separate command trees, `crank alpha` and `crank beta`, for commands that are in the alpha and beta stages of maturity. This is helpful for users in that it shows clearly which commands are non-production-ready or liable to change in breaking ways. However, it also means that moving a command to a higher maturity level is itself a breaking change: any users depending on `crank beta trace` will be broken if/when it becomes just `crank trace`.

I propose that we stop using command prefixes to denote maturity, and instead encode maturity into the commands themselves. Specifically:

- Introduce a mechanism to mark a command as alpha/beta using a kong struct tag.
- Hide alpha and beta commands by default.
- Introduce a configuration option (and a command to set it) that enables alpha and/or beta commands.
- When alpha/beta commands are enabled, denote them as such in help text.

adamwg:

Per the DevEx design doc, reorganize the command tree to be noun-centric. As part of this, also remove the alpha and beta command trees and start indicating command maturity using kong tags. Hide alpha and beta commands by default and indicate their maturity level in help when they're visible. Don't prevent users from running the commands if they're "disabled"; just hide them.

To allow enabling alpha and beta features, introduce a config file and a `crossplane config` command to manage it. The file is simple yaml, and for now enabling alpha and beta features are the only options. Show how to do this in the help text.

To avoid breaking users, leave `crossplane render` as a hidden alias for `crossplane composition render`. All the other commands that have moved were alpha or beta, so allow their old invocations to break.

Fixes crossplane/crossplane#7309

I have: 


adamwg:
Checklist intentionally incomplete since we do need a docs issue for the breaking changes.

cmd/crossplane/composition/composition.go
```diff
@@ -0,0 +1,29 @@
...
+package composition
+
+import (
+	"github.com/crossplane/cli/v2/cmd/crossplane/convert"
+	"github.com/crossplane/cli/v2/cmd/crossplane/render/xr"
+)
+
+// Cmd contains commands for working with Crossplane Compositions.
+type Cmd struct {
+	Convert convert.Cmd `cmd:"" help:"Convert a Composition to a newer version or kind." maturity:"beta"`
```
haarchri: what is or kind ?

internal/config/config.go
```diff
@@ -0,0 +1,127 @@
...
+// Config is the on-disk configuration for the crossplane CLI.
+type Config struct {
+	// Version is the version of the config file.
+	Version int `json:"version"`
+
+	Features Features `json:"features,omitzero"`
+}
+
+// Features configures feature visibility.
+type Features struct {
```
haarchri, line 42: so its all or nothing ?, can we have beta by default ? as crossplane core ?

haarchri (approved):
two small questions - all other looks good
please take care of the docs updates

cmd/crossplane/composition/composition.go
adamwg: Updated to just "newer version".

internal/config/config.go
adamwg: Yep, all or nothing. Given that we're just showing/hiding the help text for the commands, I don't think it's worth making it possible to toggle each individual one.

PR #5, opened 2026-05-07
Use function protobufs from function-sdk-go

adamwg:

We use the function protobufs in the function we implement for injecting and extracting context during render. When this code lived in crossplane/crossplane it made sense to use the protobufs from there, but now that we're separate it would be better to use the versions in function-sdk-go like a normal function.

This removes our dependency on crossplane/crossplane, which is nice.

I have: 


haarchri (approved):
any downsides ? negatives impacts around updating as soon as possible the sdks after crossplane updates ?

adamwg:
> any downsides ? negatives impacts around updating as soon as possible the sdks after crossplane updates ?

Only downside I can see is that we'll have to update function-sdk-go to pull in any new changes in Crossplane. Any changes should be backward compatible, though, so I don't expect we'll require the latest features quickly given our usage.

PR #6, opened 2026-05-08
fix(config): Fix flag parsing with space-separated value

tampakrap:

While reviewing/testing #3 I noticed that the space-separated `--config /path/to/config.yaml` ignores the config. Reproducer:

```bash
# Create a new config file
go run ./cmd/crossplane --config=/tmp/xpcfg.yaml config set features.enableAlpha true

# Correctly prints the full content of the config
go run ./cmd/crossplane --config=/tmp/xpcfg.yaml config view

# (before the fix) prints only "version: 1" ignoring the config
go run ./cmd/crossplane --config /tmp/xpcfg.yaml config view
```

The problem is that `strings.TrimPrefix` returns the original string unchanged when the prefix doesn't match, so the `v != ""` check can not distinguish "prefix matched and value extracted" from "prefix not matched". For `--config /path` this makes `configFlag` return "--config" as the path instead of falling through to read the next argument.

Switching to `strings.CutPrefix`, which returns an ok bool that makes the distinction clear.

I have: 

- [ ] ~Added or updated unit tests.~

### DIFF
--- a/cmd/crossplane/main.go
+++ b/cmd/crossplane/main.go
@@ -142,6 +142,9 @@ func configFlag(args []string) (string, error) {
 		}
 
 		if v, ok := strings.CutPrefix(a, "--config="); ok {
+			if v == "" {
+				return "", errors.New("flag --config requires a value")
+			}
 			return v, nil
 		}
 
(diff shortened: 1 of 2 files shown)


adamwg (approved):
Good catch, thanks for fixing this. Might be worth adding a test as well.

adamwg (approved):
lgtm, thanks for the test.

PR #8, opened 2026-05-10
feat(claimtoxr): Add new subcommand `resource convert claim-to-xr`

tampakrap:

Introduce a Claim to XR converter: `crossplane resource convert claim-to-xr`. Similarly to other converter functions/tools, it takes a Claim either on stdin or as file argument and generates an equivalent XR YAML.

The parent `crossplane resource convert` is marked as maturity "alpha", which applies to the subtree as well.

This new tool can be used together with the render command as shown in the example below:

```bash
crossplane render <(crossplane resource convert claim-to-xr claim.yaml) composition.yaml functions.yaml
```

The library function ConvertClaimToXR can be used by downstream tools (eg crossplane-diff) that rely on Claim to XR conversion.

The tool (and as a result the converter function, via the Options struct) is configurable and provides the following flags:
- `--direct`: omit spec.claimRef and the equivalent labels.
- `--name`: specify an explicit XR name, skipping either the random suffix in the non-direct mode or the Claim name fallback in direct mode.
- `--kind`: override the derived "X<Kind>" default.
- `--gen-uid`: populate metadata.uid with a fresh random UUID

I have: 


tampakrap:
I have created a branch for `crossplane-diff` in order to showcase how it would look like when using this tool https://github.com/crossplane-contrib/crossplane-diff/compare/main...tampakrap:crossplane-diff:theo/use_convertclaimtoxr_from_xprin?expand=1

(I am using `xprin-helpers convert-claim-to-xr` which is identical, since the crossplane/cli is not a dep of diff yet and it will require some work to do that)

tampakrap:
I chose `crossplane resource convert claim-to-xr` as I plan to introduce in a followup PR `crossplane resource patch-xr`. I am open to suggestions though.

cmd/crossplane/resource/convert/claimtoxr/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Cmd arguments and flags for converting a Crossplane Claim to an XR (Composite Resource).
+type Cmd struct {
+	// Arguments.
+	InputFile string `arg:"" default:"-" help:"The Claim YAML file to be converted. If not specified or '-', stdin will be used." optional:"" predictor:"file" type:"path"`
+
+	// Flags.
+	OutputFile string `help:"The file to write the generated XR YAML to. If not specified, stdout will be used."                                                      placeholder:"PATH" predictor:"file" short:"o" type:"path"`
+	Name       string `help:"The name to use for the XR. If empty, defaults to the Claim's name (direct mode) or the Claim's name with a random suffix (non-direct)." placeholder:"NAME" type:"string"`
+	Kind       string `help:"The kind to use for the XR. If not specified, 'X' will be prepended to the Claim's kind (e.g. Infra -> XInfra)."                         placeholder:"KIND" type:"string"`
```
haarchri, line 42: for me it feels safer to have the corresponding XRD as input - so i don't need to guess the Kind

cmd/crossplane/resource/convert/claimtoxr/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			return errors.Wrap(err, "Unable to open output file")
+		}
+
+		defer func() { _ = f.Close() }()
+
+		output = f
+	}
+
+	outputW := bufio.NewWriter(output)
```
haarchri, line 133: compared to #7  we should use `afero.WriteFile(...)`

cmd/crossplane/resource/convert/claimtoxr/converter.go
```diff
@@ -0,0 +1,175 @@
...
+	//   - true:  no spec.claimRef; no claim-name/claim-namespace labels
+	//   - false: spec.claimRef is set; claim-name/claim-namespace labels added
+	Direct bool
+
+	// GenerateUID, when true, sets metadata.uid to a fresh random UUID.
+	GenerateUID bool
+}
+
+// ConvertClaimToXR converts a Crossplane Claim to a Composite Resource (XR).
+func ConvertClaimToXR(claim *unstructured.Unstructured, opts Options) (*composite.Unstructured, error) {
```
haarchri, line 66: do we need this to be exported ?, yes if we would switch it to internal folder - and thats what i would do

cmd/crossplane/resource/resource.go
```diff
@@ -18,12 +18,14 @@ limitations under the License.
...
 
 import (
+	"github.com/crossplane/cli/v2/cmd/crossplane/resource/convert"
 	"github.com/crossplane/cli/v2/cmd/crossplane/trace"
 	"github.com/crossplane/cli/v2/cmd/crossplane/validate"
 )
 
 // Cmd contains commands for working with Crossplane resources.
 type Cmd struct {
+	Convert  convert.Cmd  `cmd:"" help:"Convert a Crossplane resource to a different kind." maturity:"alpha"`
```
haarchri, line 28:
```suggestion
	Convert  convert.Cmd  `cmd:"" help:"Convert a Crossplane resource to a different kind (e.g. Claim to XR)." maturity:"alpha"`
```

cmd/crossplane/resource/convert/convert.go
```diff
@@ -0,0 +1,42 @@
...
+// Help returns help message for the convert command.
+func (c *Cmd) Help() string {
+	return `
+This command converts a Crossplane resource to a different kind.
+
+Currently supported conversions:
+  * Claim -> Composite Resource (XR)
+
+Examples:
+  # Convert a Claim YAML file to an XR YAML file.
```
haarchri, line 39: can we add a real example for what this command is helpful?

```
  # Render a Composition using a converted Claim as the XR input.
  crossplane render <(crossplane resource convert claim-to-xr claim.yaml) composition.yaml functions.yaml
```

cmd/crossplane/resource/convert/claimtoxr/converter.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if err := xrPaved.SetString("apiVersion", apiVersion); err != nil {
+		return nil, errors.Wrap(err, "failed to set apiVersion")
+	}
+
+	// Set XR kind - either from opts or by prepending X to Claim's kind
+	kind := opts.Kind
+	if kind == "" {
+		kind = "X" + claimKind
```
haarchri, line 119: lets check my comment above regarding XRD input

haarchri:
@tampakrap regarding your comment:

> I chose crossplane resource convert claim-to-xr as I plan to introduce in a followup PR crossplane resource patch-xr. I am open to suggestions though.

i would lean toward grouping these under xr as the noun:
```
crossplane xr generate claim.yaml -o xr.yaml
crossplane xr patch xr.yaml ...

crossplane render xr.yaml composition.yaml function.yaml
crossplane render <(crossplane xr generate claim.yaml) composition.yaml functions.yaml

```


- XR is the first-class object in v2, making it the noun matches the mental model 
- Mirrors the noun-grouped pattern we already use since changed
- No hyphenated subcommands

Wdyt?

cmd/crossplane/resource/convert/claimtoxr/converter.go
jcogilvie: I think we want this exposed as an API so downstream folks (e.g. diff) can use it w/o having to invoke the CLI.  (See the [example](https://github.com/crossplane-contrib/crossplane-diff/compare/main...tampakrap:crossplane-diff:theo/use_convertclaimtoxr_from_xprin?expand=1))

cmd/crossplane/resource/convert/claimtoxr/cmd.go
tampakrap: good idea, I'd prefer to do that in a followup PR though, because I want to introduce `xr patch --xrd` and there will be a bit of refactoring

cmd/crossplane/resource/convert/claimtoxr/cmd.go
tampakrap: addressed in #13

cmd/crossplane/resource/convert/claimtoxr/converter.go
tampakrap: correct, it is mentioned as well in the description of the PR

cmd/crossplane/resource/resource.go
tampakrap: didn't do that because the command is moved to `xr generate`

cmd/crossplane/resource/convert/convert.go
tampakrap: added in #13

PR #9, opened 2026-05-10
feat(xrd): Add new subcommand `crossplane xrd convert`

tampakrap:

Introduce an XRD to CRD converter: `crossplane xrd convert`. It takes an XRD either on stdin or as a file argument and emits the equivalent Kubernetes CustomResourceDefinition that Crossplane derives from it.

The output CRD is what Crossplane uses internally to validate composite resources. This is useful for inspecting the CRD shape, feeding it into kubectl-based tooling that doesn't understand XRDs, or for debugging purposes.

The top-level `crossplane xrd` command is marked as maturity "beta", which applies to the subtree as well.

I have: 

- [ ] ~Added or updated unit tests.~


cmd/crossplane/xrd/convert.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			return errors.Wrap(err, "cannot open output file")
+		}
+
+		defer func() { _ = f.Close() }()
+
+		output = f
+	}
+
+	outputW := bufio.NewWriter(output)
```
haarchri: why Not using here `afero.WriteFile` ?

cmd/crossplane/xrd/convert.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// ToCRD converts a Crossplane XRD into a Kubernetes CRD that describes the
+// Composite Resource type, ready to be serialized. The returned CRD's
+// TypeMeta is populated so YAML/JSON marshaling produces a well-formed
+// `kind: CustomResourceDefinition` document, which is something that
+// the underlying xcrd.ForCompositeResource helper does not do on its own.
+// Callers that consume the CRD in-memory should call xcrd.ForCompositeResource directly.
+func ToCRD(xrd *apiextensionsv1.CompositeResourceDefinition) (*extv1.CustomResourceDefinition, error) {
+	crd, err := xcrd.ForCompositeResource(xrd)
```
haarchri: for legacy (V1) we also have Claim CRDs supported, so we need to handle the case to generate 2 CRDs ?
via:
`claimCrd, err = xcrd.ForCompositeResourceClaim(&xrd)`

haarchri: need to Check for claimNames e.g

haarchri (commented):
thanks, can you add also unit-tests for the convert command? xrd for namespace scope, Cluster scope and legacy with / without claimNames

cmd/crossplane/xrd/convert.go
```diff
@@ -0,0 +1,144 @@
...
+
+	commonIO "github.com/crossplane/cli/v2/cmd/crossplane/convert/io"
+)
+
+type convertCmd struct {
+	// Arguments.
+	InputFile string `arg:"" default:"-" help:"The XRD YAML file to be converted. If not specified or '-', stdin will be used." optional:"" predictor:"file" type:"path"`
+
+	// Flags.
+	OutputFile string `help:"The file to write the generated CRD YAML to. If not specified, stdout will be used." placeholder:"PATH" predictor:"file" short:"o" type:"path"`
```
haarchri: wonder if it's better to have an `OutputFolder` as we can have multiple CRDs

cmd/crossplane/xrd/convert.go
tampakrap: I copy-pasted it from the older/existing converters eg https://github.com/crossplane/cli/blob/eb4d56a1a40cc66c728b7a8884fb914dc69d951a/cmd/crossplane/convert/compositionenvironment/cmd.go#L120 but I don't mind changing it, what would you prefer?

cmd/crossplane/xrd/convert.go
tampakrap: another idea would be to introduce a --legacy flag

haarchri: feels not obious when i have a V1 or V2 XRD and want to convert, as crossplane is doing it automatically, i would expect it here too without extra flags

tampakrap: I added both using the kong's `xor` option

cmd/crossplane/xrd/convert.go
tampakrap: done, together with unit tests on a separate commit, please give me another review

cmd/crossplane/xrd/convert.go
haarchri: regarding the implementation for devex features i would tent to using afero.WriteFile - as we interacting with an aferoFS anyways - #7

cmd/crossplane/xrd/convert.go
```diff
@@ -0,0 +1,206 @@
...
+CustomResourceDefinition (CRD) that describes the composite resource type.
+
+The output CRD(s) are what Crossplane derives internally from the XRD. This is
+useful for inspecting the CRD shape, feeding it into kubectl-based tooling that
+doesn't understand XRDs, or as a debugging aid.
+
+For legacy XRDs that offer a Claim (spec.claimNames set, typically with
+spec.scope: LegacyCluster) two CRDs are produced: one cluster-scoped CRD for
+the XR and one namespaced CRD for the Claim. For namespaced XRDs only the XR
+CRD is produced. The detection is automatic; no flag is needed.
```
haarchri, line 63: can we make this here a bit clearer ?
```suggestion
Convert a CompositeResourceDefinition (XRD) into the CustomResourceDefinition(s)
that Crossplane derives from it internally.

Useful for inspecting the generated CRD shape, feeding it into kubectl-based
tooling that doesn't understand XRDs, or debugging composition behavior.

Output depends on the XRD type, detected automatically:
  * Namespaced XRD             -> one CRD for the XR
  * Legacy XRD with a Claim    -> two CRDs: cluster-scoped XR + namespaced Claim
```

cmd/crossplane/xrd/convert.go
```diff
@@ -0,0 +1,206 @@
...
+func writeCRDs(output io.Writer, crds []*extv1.CustomResourceDefinition) error {
+	outputW := bufio.NewWriter(output)
+
+	for _, crd := range crds {
+		b, err := yaml.Marshal(crd)
+		if err != nil {
+			return errors.Wrapf(err, "cannot marshal CRD %q", crd.GetName())
+		}
+
+		if _, err := outputW.WriteString("---\n"); err != nil {
```
haarchri, line 152: my comment is still valid - can we use here afero.WriteFile like we doing in #7

cmd/crossplane/xrd/convert.go
```diff
@@ -0,0 +1,206 @@
...
+// `kind: CustomResourceDefinition` documents, which is something that the
+// underlying xcrd helpers do not do on their own.
+//
+// For legacy XRDs that offer a Claim the result is a two-element slice:
+// the CRD for the XR followed by the CRD for the Claim. For namespaced XRDs
+// the result is a single-element slice containing only the XR CRD.
+//
+// Callers that consume the CRD in-memory should call xcrd.ForCompositeResource
+// (and, for legacy XRDs, xcrd.ForCompositeResourceClaim) directly.
+func ToCRDs(xrd *apiextensionsv1.CompositeResourceDefinition) ([]*extv1.CustomResourceDefinition, error) {
```
haarchri, line 180: is this something we need to export ? wonder if this is something we can move to Internal

cmd/crossplane/xrd/convert.go
tampakrap: addressed in #12

cmd/crossplane/xrd/convert.go
tampakrap: addressed in #12

cmd/crossplane/xrd/convert.go
tampakrap: addressed in #12

cmd/crossplane/xrd/convert.go
tampakrap: I agree that it doesn't need to be exported, but I didn't move it to internal since no other package needs it, at least for now. If any other package needs it in the future we can move it to internal. The change is done in #12

PR #10, opened 2026-05-11
Introduce an initial set of DevEx features

### LINKED ISSUE #6840: Proposal: Contribute the Upbound developer experience tooling to Crossplane

The difficulty of building compositions is a hurdle to Crossplane adoption for many users. The original patch-and-transform composition system was easy to get started with but became unwieldy for large compositions. The modern function pipeline style of composition provides more power and readability by allowing for the use of general purpose programming languages; however, users either have to build and manage composition functions separately from their compositions, or use "generic functions" and write composition code inline in their compositions, neither of which is a simple experience.

Upbound has built a developer experience centered on the idea of control plane "projects": an on-disk directory layout that includes both Crossplane resources (XRDs and compositions in YAML files) and functions. Projects are built into a set of Crossplane packages, with dependencies between them, by the Upbound CLI. There is also support for running a local control plane for testing, and for writing automated tests (both render-based "composition tests" and uptest-based e2e tests).

Full details can be found in the Upbound documentation: https://docs.upbound.io/manuals/cli/howtos/project/

Upbound would like to contribute this tooling to the Crossplane project, either as part of the existing crossplane/crank CLI, or as a new set of tools. Specifically, we propose porting the following functionality from the proprietary `up` CLI to an open source Crossplane tool:

* The "project" concept and API definitions.
* Commands for managing projects and their dependencies.
* Commands for authoring XRDs and compositions within a project.
* Commands to "compile" projects, including their embedded functions, into Crossplane packages, and pushing those packages to a registry.
* The ability to easily run a control plane locally and install a project into it for testing.
* Commands for testing projects.
* Tooling for generating and distributing "schemas" (bindings for various programming langauges) for Crossplane providers and configurations.

This issue is intended as a starting point for discussion. A design doc describing the details of where this functionality will land in Crossplane and how it works will follow.

adamwg:

This PR introduces the initial set of DevEx tools based around the concept of control plane Projects. Specifically, it introduces tools for:

- Scaffolding a new, empty project.
- Generating XRDs from example XRs or simpleschema documents.
- Scaffolding compositions.
- Scaffolding functions in go, go-templating, kcl, and python and adding them to composition pipelines.
- Managing both runtime and build-time dependencies for a project, including generating language bindings (schemas) for added dependencies.
- Building a project into a set of xpkgs.
- Installing a project into a local control plane (created using kind) for testing.
- Support for projects and embedded functions in `crossplane composition render` and `crossplane operation render`.

Fixes crossplane/crossplane#6840

I have: 


cmd/crossplane/function/templates/python/function/fn.py
```diff
@@ -0,0 +1,29 @@
+"""A Crossplane composition function."""
+
+import grpc
+from crossplane.function import logging, response
+from crossplane.function.proto.v1 import run_function_pb2 as fnv1
+from crossplane.function.proto.v1 import run_function_pb2_grpc as grpcv1
+
```
haarchri, line 7: with this version we doing this without "schemas" ? but i can see schemas generator - so how does this differ from the up-cli ? do we still need . in front =

internal/terminal/spinner.go
```diff
@@ -0,0 +1,447 @@
...
+
+	bspinner "github.com/charmbracelet/bubbles/spinner"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+
+	"github.com/crossplane/cli/v2/internal/async"
+)
+
+var (
+	// Crossplane teal for dark backgrounds, dark blue for light backgrounds.
```
haarchri, line 38: are these the real crossplane colors ?

cmd/crossplane/render/op/cmd.go
```diff
@@ -58,7 +71,10 @@ type Cmd struct {
 	RequiredSchemas        string            `help:"A directory of JSON files specifying OpenAPI schemas to pass to the function pipeline."                                                    placeholder:"DIR"       predictor:"directory"              type:"path"`
 	WatchedResource        string            `help:"A YAML file specifying the watched resource for WatchOperation rendering. The resource is also added to required resources."               placeholder:"PATH"      predictor:"yaml_file"              short:"w"   type:"existingfile"`
 
-	Timeout time.Duration `default:"1m" help:"How long to run before timing out."`
+	CacheDir       string        `env:"CROSSPLANE_XPKG_CACHE"       help:"Directory for cached xpkg package contents."          name:"cache-dir"`
+	MaxConcurrency uint          `default:"8"                       help:"Maximum concurrency for building embedded functions."`
+	ProjectFile    string        `default:"crossplane-project.yaml" help:"Path to the project file."                            short:"f"        type:"path"`
```
haarchri: does this mean everybody needs to have a "project" that render is working for op ?

apis/dev/v1alpha1/defaults.go
```diff
@@ -0,0 +1,51 @@
...
+		p.Examples = "examples"
+	}
+	if p.Tests == "" {
+		p.Tests = "tests"
+	}
+	if p.Operations == "" {
+		p.Operations = "operations"
+	}
+	if p.Schemas == "" {
+		p.Schemas = "schemas"
```
haarchri, line 49: this is for what? we not allowing apiDependencies and dependsOn as input for schemas ? that we can render ?

cmd/crossplane/function/templates/python/function/fn.py
adamwg: Three changes from the `up` version:

1. Schemas are always generated client-side rather than packaged as xpkg layers. We intend for users to check the `schemas/` dir into git with their project, and the lockfile ensures schemas aren't regenerated unless deps change.
2. We use full Python projects rather than the simplified Python experience we had in `up`. This means the `pyproject.toml` can reference the schemas as `"crossplane-models @ file:./../../schemas/python"` and they get pulled in from the schema dir as part of the build process, eliminating the need for the symlink and leading `.`.
3. To make the imports nicer with the above, we put all the generated Python schemas under `models/`, so that they're imported like `models.io.k8s.api.apps`.

I'll share an example project shortly with a Python function to show how everything looks.

apis/dev/v1alpha1/project_types.go
```diff
@@ -0,0 +1,188 @@
...
+	// order for dependencies on embedded functions to resolve correctly.
+	Repository string `json:"repository"`
+
+	// Crossplane defines the Crossplane version constraints for the
+	// configuration package built from the project. If not specified, the
+	// constraint will be '>=v2.0.0-rc.0' such that the packages support any
+	// Crossplane 2.x release.
+	Crossplane *pkgmetav1.CrossplaneConstraints `json:"crossplane,omitempty"`
+	// Dependencies are built-time and runtime dependencies of the project.
+	Dependencies []Dependency `json:"dependencies,omitempty"`
```
haarchri, line 66: this is old dependsOn ?

internal/terminal/spinner.go
adamwg: Yes, colors taken from https://www.lingoapp.com/88697/s/Colors-o625M5?v=0

apis/dev/v1alpha1/project_types.go
```diff
@@ -0,0 +1,188 @@
...
+	// development. Currently, only rewriting is supported; other options will
+	// be silently ignored. Note that these configs are for development only;
+	// any necessary ImageConfigs for deployment into a cluster must be created
+	// separately at deployment time.
+	ImageConfigs []pkgv1beta1.ImageConfig `json:"imageConfigs,omitempty"`
+}
+
+// ProjectPackageMetadata holds metadata about the project, which will become
+// package metadata when a project is built into a Crossplane package.
+type ProjectPackageMetadata struct {
```
haarchri, line 82: is this up2date with "marketplace" guys - or we decouple this here ?

apis/dev/v1alpha1/project_types.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Version is the semantic version constraints for the dependency.
+	Version string `json:"version"`
+
+	// APIOnly indicates that this dependency is only needed for API/schema
+	// purposes and should not be included as a runtime dependency in the
+	// built package. Only xpkg dependencies can be runtime dependencies.
+	// Default is false, meaning xpkg dependencies are runtime by default.
+	// +optional
+	APIOnly bool `json:"apiOnly,omitempty"`
```
haarchri, line 161: do we have an case that an xpkgDependency is not a dependency? - like Configurations skipping dependency ?

haarchri:
@adamwg can you provide some demo/template repositories based on this PR ? this makes it much easier to review

cmd/crossplane/function/templates/python/function/fn.py
haarchri: means all good with schemas taking some times ? do we have some docs regarding this ? what we expect here for something like ec2 provider?, compute provider etc ? what happen if we bring beack monolith ? do we have options for MRAPs in project ?

adamwg: Right, for now we'll accept that adding a dep takes a while. For reference, on my machine `upbound/provider-aws-ec2` takes about 2 minutes and `upbound/provider-gcp-compute` takes about 3 minutes.

I do think we should optimize this in the future. A few possible (not mutually exclusive) options:

- Package pre-generated schemas with packages, like we did in `up` (though maybe as related OCI artifacts rather than layers, so k8s doesn't have to pull them) and avoid generating client-side when these are present.
- Define a way to find pre-generated schemas in standard language-specific repositories (e.g., a mapping between xpkgs and pip modules for python).
- Optimize client-side generation by using MRAPs (or a more generic equivalent) to generate schemas for a subset of resources.
- Add a `langauges:` field to the Project spec and avoid generating schemas for languages the user doesn't care about. I believe Python schemas are the slowest by far, so this would make things a lot faster for non-Python users.

cmd/crossplane/render/op/cmd.go
adamwg: No, `render` will work as it does today if the users provides a functions file and/or is not in a project directory (functions file is required when not in a project directory).

I'll update the `ProjectFile` arg here to be `optional:""` so this is clearer.

apis/dev/v1alpha1/defaults.go
adamwg: This is the directory into which schemas will be generated (and where the function builders will look for schemas) - like the `.up` directory in the Upbound DevEx. Figured we may as well make it configurable rather than hard-coding to `schemas/`.

apis/dev/v1alpha1/project_types.go
adamwg: Correct, consolidated `dependsOn` and `apiDependencies` into a single set based on feedback in the design doc.

apis/dev/v1alpha1/project_types.go
adamwg: With `apiOnly: true` we'll generate schemas for the xpkg, but not add it to the `dependsOn` of the configuration. The only realistic use-case I can think of for this right now is if a user wants to build with provider schemas but install providers manually rather than using the package manager.

adamwg:
> @adamwg can you provide some demo/template repositories based on this PR ? this makes it much easier to review

Yep, good idea. Building some examples here: https://github.com/adamwg/crossplane-devex-examples (only Python so far, will work on others).

I will also update https://github.com/adamwg/kubecon-eu-2026-devex-demo to reflect the current state of the tooling so it can be used as an end-to-end testing script.

internal/project/functions/go_templating.go
```diff
@@ -0,0 +1,159 @@
...
+		matches = true
+		return nil
+	})
+
+	if errors.Is(err, fs.SkipAll) {
+		err = nil
+	}
+
+	return matches, err
+}
```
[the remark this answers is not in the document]
adamwg: This is intentional. We return `true` only when *all* files in the directory are regular and have go templating file extensions.

internal/project/build.go
```diff
@@ -0,0 +1,600 @@
...
+				return errors.Wrapf(err, "failed to construct index for function image %q", fnName)
+			}
+			dgst, err := idx.Digest()
+			if err != nil {
+				return errors.Wrapf(err, "failed to get index digest for function image %q", fnName)
+			}
+			deps[i] = xpmetav1.Dependency{
+				APIVersion: new(xpkgv1.FunctionGroupVersionKind.GroupVersion().String()),
+				Kind:       new(xpkgv1.FunctionKind),
+				Package:    &fnRepo,
```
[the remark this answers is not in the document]
adamwg: `new` to create a pointer to an arbitrary value is a new feature in Go 1.26.

apis/dev/v1alpha1/project_types.go
adamwg: In `up` we've added `AdditionalMetadata`, which can hold arbitrary k/v pairs that are specifically used by the Upbound marketplace (and must be prefixed with `meta.upbound.io/`). I've left those out here for now since they're pretty Upbound-specific and the k/v pairs get copied directly into the resulting configuration package's `metadata.annotations` so it's possible to fill them in just by writing them directly in the project's `metadata.annotations`. We can always add them in the future if desired.

cmd/crossplane/dependency/dependency.go
```diff
@@ -0,0 +1,25 @@
...
+
+// Package dependency contains commands for managing project dependencies.
+package dependency
+
+// Cmd contains commands for dependency management.
+type Cmd struct {
+	Add         addCmd         `cmd:"" help:"Add a dependency to the current project."`
+	UpdateCache updateCacheCmd `cmd:"" help:"Update the dependency cache for the current project."`
+	CleanCache  cleanCacheCmd  `cmd:"" help:"Clean the dependency cache."`
+}
```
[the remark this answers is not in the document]
adamwg: I'll update the coderabbit config for this later - we don't have feature flags in the CLI, instead using the `maturity` tag on commands (which these commands have on their parent).

internal/kcl/import.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Resolve collisions by adding more context from earlier components.
+	if existingAliases[alias] {
+		for i := 3; i <= len(parts); i++ {
+			alias = strings.Join(parts[len(parts)-i:], "")
+			if !existingAliases[alias] {
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
adamwg: The only way the alias can be non-unique at the end of this loop is if there are two identical import paths (i.e., the same module is imported twice). Given that we generate the set of import paths based on which directory paths exist in the schemas directory, we'll never have the same module path twice.

internal/schemas/generator/go.go
```diff
@@ -0,0 +1,1365 @@
...
+		// remove the self‑import (e.g. meta/v1 importing itself)
+		code, err = removeSelfImports(code, pkg)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := writeGoCode(schemaFS, group, kind, version, code); err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
adamwg: Unclear to me whether this is a real issue. The code here matches what we have in `up`, but I know we've had some bugs there recently with complex CRDs.

@haarchri You may have more context than I do on this part - any thoughts on whether coderabbit's issue is something we need to fix?

cmd/crossplane/render/op/cmd.go
adamwg: Updated the help to make it clearer that the project file is optional, and the functions file can be omitted when using render in a project.

internal/kcl/import.go
```diff
@@ -0,0 +1,76 @@
...
+
+	existingAliases[alias] = true
```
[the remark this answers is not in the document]
adamwg: Rewrote the function so it takes a list of paths and returns an alias -> path mapping. That lets us keep the existing paths internal to the function rather than passing them in iteratively, and cleans up the calling code.

internal/project/functions/go.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (b *goBuilder) match(fromFS afero.Fs) (bool, error) {
+	return afero.Exists(fromFS, "go.mod")
+}
+
+func (b *goBuilder) Build(ctx context.Context, c BuildContext) ([]v1.Image, error) {
+	// ko logs using the Go standard library global logger without providing
+	// any option to disable output. Disable output while we do our builds.
+	log.SetOutput(io.Discard)
```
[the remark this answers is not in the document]
adamwg: I don't think we ever expect output from the standard logger, but it doesn't hurt to restore it when we return.

internal/schemas/generator/go.go
```diff
@@ -0,0 +1,1365 @@
...
+func goRenameTypes(s *spec3.OpenAPI) {
+	for name, schema := range s.Components.Schemas {
+		goName := goFixName(name)
+		if goName == "" {
+			delete(s.Components.Schemas, name)
+		}
+		goRenameSchemaType(goName, schema)
+		goRenamePropertyTypes(goName, schema.Properties)
+	}
+}
```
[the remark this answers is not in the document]
adamwg: I think this analysis is correct - it's harmless to mutate the `schema` after removing it from the map, but also unnecessary. Added the `continue` as suggested.

internal/docker/docker.go
```diff
@@ -467,6 +474,77 @@
...
+		return errors.Wrap(err, "failed to copy files from container")
+	}
+
+	tarReader := tar.NewReader(reader)
+
+	// Limit files to 1GiB to avoid excessive memory usage.
+	const maxFileSize = 1024 * 1024 * 1024
+
+	for {
+		header, err := tarReader.Next()
```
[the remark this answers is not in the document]
adamwg: Updated to reject relative paths that leave the `path` prefix.

internal/project/functions/build_test.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+		})
+	}
+}
+
+//go:embed testdata/kcl-function/**
+var kclFunction embed.FS
+
+func TestKCLBuild(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
adamwg: Updated the static config store to be thread-safe by giving each one its own scheme.

cmd/crossplane/render/xr/cmd.go
```diff
@@ -63,8 +76,11 @@ type Cmd struct {
...
 
-	Timeout time.Duration `default:"1m"                                                                                                     help:"How long to run before timing out."`
-	XRD     string        `help:"A YAML file specifying the CompositeResourceDefinition (XRD) that defines the XR's schema and properties." optional:""                               placeholder:"PATH" type:"existingfile"`
+	CacheDir       string        `env:"CROSSPLANE_XPKG_CACHE"                                                                                      help:"Directory for cached xpkg package contents."          name:"cache-dir"`
+	MaxConcurrency uint          `default:"8"                                                                                                      help:"Maximum concurrency for building embedded functions."`
+	ProjectFile    string        `default:"crossplane-project.yaml"                                                                                help:"Path to the project file. Optional."                  optional:""        predictor:"yaml_file" short:"f" type:"existingfile"`
```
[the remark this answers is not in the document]
adamwg: Good catch, fixed.

haarchri (approved):

adamwg:
Force pushed to fix conflicts and add markdown-formatted help for the new commands (since #14 is merged).

cmd/crossplane/function/generate.go
```diff
@@ -0,0 +1,410 @@
...
+	ImportPath string
+	Alias      string
+}
+
+func (c *generateCmd) generateKCLFiles(fs afero.Fs) error {
+	tmpls := template.Must(template.ParseFS(kclTemplates, "templates/kcl/*"))
+
+	foundFolders, _ := filesystem.FindNestedFoldersWithPattern(c.schemasFS, "kcl", "*.k")
+
+	imports := kcl.FormatKclImportPaths(foundFolders)
```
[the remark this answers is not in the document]
adamwg: This is intentional, to handle the case where there are no schemas (e.g., because the user has not added any dependencies).

internal/docker/docker.go
```diff
@@ -467,6 +475,85 @@ func RunContainer(ctx context.Context, img string, opts ...RunContainerOption) (
...
+		}
+
+		if header.Size > maxFileSize {
+			return errors.Errorf("file %s is over 1GiB - refusing to copy", header.Name)
+		}
+
+		cleanedPath := filepath.Clean(strings.TrimPrefix(header.Name, filepath.Base(path)+"/"))
+		if slices.Contains(filepath.SplitList(cleanedPath), "..") {
+			return errors.Errorf("file %s lives outside path %s - refusing to copy", cleanedPath, path)
+		}
```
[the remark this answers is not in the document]
adamwg: Good catch! Fixed by using `strings.Split` and hard-coding `/` as the path separator  since tar files always use the Unix convention.

PR #11, opened 2026-05-12
feat(xpkg): add --annotation flag to xpkg build and xpkg push

chaitanyapantheor:

Adds a repeatable `--annotation/-a KEY=VALUE` flag to `crossplane xpkg build`
and `crossplane xpkg push`, allowing users to attach OCI manifest annotations
to packages at build or push time.

```sh
# Build with a static annotation
crossplane xpkg build -a org.opencontainers.image.source=https://github.com/org/repo

# Push with a dynamic (CI-injected) annotation
crossplane xpkg push -f pkg.xpkg \
  -a org.opencontainers.image.revision=$(git rev-parse HEAD) \
  xpkg.example.io/org/pkg:v1.0.0
```
Annotations are applied to the OCI image manifest. Malformed annotations
(missing =) return an error before any write or push occurs. Reading
annotations from crossplane.yaml is intentionally out of scope to avoid
silently propagating internal Crossplane annotations to the OCI registry.

Covered by unit tests in cmd/crossplane/xpkg/annotations_test.go.

Fixes https://github.com/crossplane/crossplane/issues/7282

I have:

- [ ] Added or updated unit tests.
- [ ] Linked a PR or a [docs tracking issue] to [document this change]. — `command-reference.md` is auto-generated from CLI help text; the fix to `build.go` is the docs update.
- [ ] Added `backport release-x.y` labels to auto-backport this PR. — `backport release-2.3` requested (needs maintainer to apply label).


cmd/crossplane/xpkg/push.go
```diff
@@ -60,8 +60,9 @@ type pushCmd struct {
 	Package string `arg:"" help:"Where to push the package. Must be a fully qualified OCI tag, including the registry, repository, and tag." placeholder:"REGISTRY/REPOSITORY:TAG"`
 
 	// Flags. Keep sorted alphabetically.
+	Annotation            []string `help:"An OCI manifest annotation to add to the package in key=value format. Repeatable." placeholder:"KEY=VALUE" short:"a"`
```
adamwg: I wonder if this should be `OCIAnnotation` / `--oci-annotation` or something like that to make it clear how it's different from the `metadata.annotations`.

cmd/crossplane/xpkg/push.go
```diff
@@ -192,6 +200,8 @@ func pushImages(logger logging.Logger, images []packageImage, url string, option
 				return errors.Wrapf(err, errAnnotateLayers)
 			}
 
+			img = annotateImage(img, annotations)
```
adamwg, line 195: Should we apply the same annotations to the index (built below with `mutate.AppendManifests`)?

cmd/crossplane/xpkg/annotations.go
```diff
@@ -0,0 +1,49 @@
+/*
+Copyright 2025 The Crossplane Authors.
```
adamwg: Nit: 2026 :-)

adamwg (commented):
I've left a couple of thoughts, but overall this looks good to me and seems like a valuable feature.

cmd/crossplane/xpkg/annotations.go
chaitanyapantheor: Fixed — updated to 2026.

cmd/crossplane/xpkg/push.go
chaitanyapantheor: Good point — renamed to `--oci-annotation` (Go field `OCIAnnotation`, explicit `name:"oci-annotation"` kong tag) in both `xpkg build` and `xpkg push`. Short flag `-a` is unchanged.

cmd/crossplane/xpkg/push.go
chaitanyapantheor: Added `annotateIndex` helper (mirrors `annotateImage`) and applied it to the index before `remote.WriteIndex` in the multi-platform path. Annotations now appear on both the individual manifests and the top-level index.

cmd/crossplane/xpkg/build.go
```diff
@@ -99,12 +100,13 @@ func (c *buildCmd) AfterApply() error {
...
-	Ignore                   []string `help:"comma-separated list of globs specifying files to exclude from the build, relative to --package-root." placeholder:"PATH"`
-	PackageFile              string   `help:"The file to write the package to. Defaults to a generated filename in --package-root."                 placeholder:"PATH"                                                     predictor:"xpkg_file" short:"o"           type:"path"`
-	PackageRoot              string   `default:"."                                                                                                  help:"The directory that contains the package's crossplane.yaml file." predictor:"directory" short:"f"           type:"existingdir"`
+	OCIAnnotation            []string `help:"An OCI manifest annotation to add to the package in key=value format. Repeatable."                                                                       name:"oci-annotation"                                                  placeholder:"KEY=VALUE" short:"a"`
```
adamwg: As far as I can tell, the image annotations actually get persisted anywhere by `crossplane xpkg build` (they're not represented anywhere in the tarball written by `tarball.Write`). Assuming that's true, we should remove the flag here and keep it on `push` only; otherwise, a user could reasonably expect that they don't need to provide annotations to `push` if they've already provided them to `build`, which isn't true.

adamwg (changes requested):
Thanks for the updates! Looks good to me, aside from the comment I've left on `build` - sorry for not calling it out earlier.

cmd/crossplane/xpkg/build.go
chaitanyapantheor: Good catch — you're right that annotations applied during build aren't stored in the tarball, so the flag was giving a false impression. I've removed `--oci-annotation` from `xpkg build` entirely; it now lives only on `xpkg push` where it can actually take effect.

adamwg (approved):
LGTM, thanks for the contribution!

chaitanyapantheor:
Maintainer note: this PR targets `backport release-2.3`. The contributor account doesn't have label permissions on this repo — could a maintainer apply the `backport release-2.3` label? Thanks!

adamwg:
> Maintainer note: this PR targets `backport release-2.3`. The contributor account doesn't have label permissions on this repo — could a maintainer apply the `backport release-2.3` label? Thanks!

Given that this PR is adding a new flag, not fixing a bug/issue, I don't think we need to backport it. The new flag will be available in the v2.4 release (which can/should happen sooner than the core Crossplane v2.4 release - that's why we moved the CLI to its own repository :-) ).

PR #13, opened 2026-05-12
feat(xr): Convert a Claim to XR via `crossplane xr generate`

tampakrap:

Introduce a Claim to XR converter: `crossplane xr generate claim.yaml`. Similarly to other converter functions/tools, it takes a Claim either on stdin or as file argument and generates an equivalent XR YAML.

The parent `crossplane xr` is marked as maturity "alpha", which applies to the subtree as well.

This new tool can be used together with the render command as shown in the example below:

```bash
crossplane render <(crossplane xr generate claim.yaml) composition.yaml functions.yaml
```

The library function ConvertClaimToXR can be used by downstream tools (eg crossplane-diff) that rely on Claim to XR conversion.

The tool (and as a result the converter function, via the Options struct) is configurable and provides the following flags:
- `--direct`: omit spec.claimRef and the equivalent labels.
- `--name`: specify an explicit XR name, skipping either the random suffix in the non-direct mode or the Claim name fallback in direct mode.
- `--kind`: override the derived "X<Kind>" default.
- `--gen-uid`: populate metadata.uid with a fresh random UUID.

I have: 

### DIFF
--- a/cmd/crossplane/xr/generate.go
+++ b/cmd/crossplane/xr/generate.go
@@ -1,0 +1,245 @@
+/*
+Copyright 2026 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package xr
+
+import (
+	"github.com/alecthomas/kong"
+	"github.com/spf13/afero"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apiserver/pkg/storage/names"
+	"sigs.k8s.io/yaml"
+
+	"github.com/crossplane/crossplane-runtime/v2/pkg/errors"
+	"github.com/crossplane/crossplane-runtime/v2/pkg/fieldpath"
+	"github.com/crossplane/crossplane-runtime/v2/pkg/resource/unstructured/composite"
+
+	commonIO "github.com/crossplane/cli/v2/cmd/crossplane/convert/io"
+
+	_ "embed"
+)
+
+//go:embed help/generate.md
+var generateHelp string
+
+type generateCmd struct {
+	// Arguments.
+	InputFile string `arg:"" default:"-" help:"The Claim YAML file to convert, or '-' for stdin." optional:"" predictor:"file" type:"path"`
+
+	// Flags.
+	OutputFile string `help:"The file to write the generated XR YAML to. Defaults to stdout."                                                                         placeholder:"PATH" predictor:"file" short:"o" type:"path"`
+	Name       string `help:"The name to use for the XR. If empty, defaults to the Claim's name (direct mode) or the Claim's name with a random suffix (non-direct)." placeholder:"NAME" type:"string"`
+	Kind       string `help:"The kind to use for the XR. Defaults to 'X' prepended to the Claim's kind (for example, Infra -> XInfra)."                               placeholder:"KIND" type:"string"`
+	Direct     bool   `help:"Create a direct XR without Claim references and suffix."                                                                                 name:"direct"      negatable:""`
+	GenUID     bool   `help:"Set a fresh random metadata.uid on the generated XR."                                                                                    name:"gen-uid"`
+
+	fs afero.Fs
+}
+
+func (c *generateCmd) Help() string {
+	return generateHelp
+}
+
+// AfterApply implements kong.AfterApply.
+func (c *generateCmd) AfterApply() error {
+	c.fs = afero.NewOsFs()
+	return nil
+}
+
+// Run runs the generate command.
+func (c *generateCmd) Run(k *kong.Context) error {
+	claimData, err := commonIO.Read(c.fs, c.InputFile)
+	if err != nil {
+		return err
+	}
+
+	claim := &unstructured.Unstructured{}
+	if err := yaml.Unmarshal(claimData, claim); err != nil {
+		return errors.Wrap(err, "cannot unmarshal Claim")
+	}
+
+	// Convert to XR
+	xr, err := ConvertClaimToXR(claim, Options{
+		Name:        c.Name,
+		Kind:        c.Kind,
+		Direct:      c.Direct,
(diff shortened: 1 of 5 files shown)


cmd/crossplane/xr/generate.go
```diff
@@ -0,0 +1,281 @@
...
+)
+
+type generateCmd struct {
+	// Arguments.
+	InputFile string `arg:"" default:"-" help:"The Claim YAML file to be converted. If not specified or '-', stdin will be used." optional:"" predictor:"file" type:"path"`
+
+	// Flags.
+	OutputFile string `help:"The file to write the generated XR YAML to. If not specified, stdout will be used."                                                      placeholder:"PATH" predictor:"file" short:"o" type:"path"`
+	Name       string `help:"The name to use for the XR. If empty, defaults to the Claim's name (direct mode) or the Claim's name with a random suffix (non-direct)." placeholder:"NAME" type:"string"`
+	Kind       string `help:"The kind to use for the XR. If not specified, 'X' will be prepended to the Claim's kind (e.g. Infra -> XInfra)."                         placeholder:"KIND" type:"string"`
```
adamwg: Instead of, or maybe in addition to, this flag, it would be nice to allow an XRD to be specified, which would provide the necessary kind mapping.

cmd/crossplane/xr/generate.go
```diff
@@ -0,0 +1,281 @@
...
+// Run runs the generate command.
+func (c *generateCmd) Run(k *kong.Context) error {
+	claimData, err := commonIO.Read(c.fs, c.InputFile)
+	if err != nil {
+		return err
+	}
+
+	claim := &unstructured.Unstructured{}
+	if err := yaml.Unmarshal(claimData, claim); err != nil {
+		return errors.Wrap(err, "Unmarshalling Error")
```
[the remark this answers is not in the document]
adamwg: I agree a bit with coderabbit on adding a little detail here, but more importantly by convention error messages should start with a lower-case letter.

```suggestion
		return errors.Wrap(err, "cannot unmarshal claim")
```

cmd/crossplane/xr/generate.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if _, err := k.Stdout.Write(data); err != nil {
+		return errors.Wrap(err, "cannot write output")
+	}
+
+	return nil
+}
+
+const (
+	// Error messages.
```
adamwg: The Crossplane style guidelines no longer suggest error message constants, preferring literals in `errors.Wrap` calls.

cmd/crossplane/xr/generate.go
```diff
@@ -0,0 +1,281 @@
...
+	errNilInput        = "input is nil"
+	errEmptyClaimYAML  = "invalid Claim YAML: parsed object is empty"
+	errNoAPIVersion    = "Claim has no apiVersion"
+	errParseAPIVersion = "failed to parse Claim APIVersion"
+	errNoKind          = "Claim has no kind section"
+	errNoSpecSection   = "Claim has no spec section"
```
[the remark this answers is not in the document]
adamwg: I disagree with coderabbit here - its suggested error messages are excessive 😆.

cmd/crossplane/xr/generate.go
```diff
@@ -0,0 +1,281 @@
...
+	//   - true:  no spec.claimRef; no claim-name/claim-namespace labels
+	//   - false: spec.claimRef is set; claim-name/claim-namespace labels added
+	Direct bool
+
+	// GenerateUID, when true, sets metadata.uid to a fresh random UUID.
+	GenerateUID bool
+}
+
+// ConvertClaimToXR converts a Crossplane Claim to a Composite Resource (XR).
+func ConvertClaimToXR(claim *unstructured.Unstructured, opts Options) (*composite.Unstructured, error) {
```
adamwg, line 136: This mostly duplicates functionality from Crossplane's claim reconciler. Could we factor some code out of the reconciler into crossplane-runtime and consume it here? Similar to how the `xcrd` package handles XRD->CRD conversion for both the CLI and the XRD reconciler.

Not against having this in the CLI for now - we can get this merged and refactor afterwards.

adamwg (commented):
LGTM overall - left a few small comments.

Thinking to the future, it would be nice to also be able to generate a scaffold/minimal XR from an XRD (we have this in `up` as `up example generate`). Maybe we could add a `--from=<claim|xrd>` flag to this command, similar to how we have `--from=<xr|simpleschema>` on `crossplane xrd generate`. No need to add it now - just thinking of whether we can add in the future without breaking the current functionality.

cmd/crossplane/xr/generate.go
tampakrap: Christopher suggested the same in the duplicate https://github.com/crossplane/cli/pull/8#discussion_r3218423389 I'll duplicate my answer here: good idea, I'd prefer to do that in a followup PR though, because I want to introduce xr patch --xrd and there will be a bit of refactoring

cmd/crossplane/xr/generate.go
tampakrap: when I initially developed this subcommand in `xprin-helpers` there was some issue or some limitation. I don't really remember what it was though, and it might not be relevant any more, so I'm definitely interested to try that once again in a followup PR

cmd/crossplane/xr/generate.go
adamwg: Thanks for the context - I missed the original discussion. Fine with saving this for a follow-up :+1:

cmd/crossplane/xr/generate.go
```diff
@@ -0,0 +1,281 @@
...
+		labels[labelClaimNamespace] = claim.GetNamespace()
+		if err := xrPaved.SetValue("spec.claimRef", map[string]any{
+			"apiVersion": apiVersion,
+			"kind":       claimKind,
+			"name":       claimName,
+			"namespace":  claim.GetNamespace(),
+		}); err != nil {
+			return nil, errors.Wrap(err, "failed to set claimRef")
+		}
+	}
```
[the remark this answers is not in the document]
tampakrap: Claim without namespace means that it will use the default namespace

tampakrap:
I chose to force-pushed to resolve the conflict, the changes that I did:
- added the markdown file with the help
- got rid of the constants with the error messages
- lowercase start for error messages
- fixed the `vale` issues

@adamwg @haarchri give me another review please

cmd/crossplane/xr/generate.go
tampakrap: fixed

adamwg (approved):
LGTM, with the follow-up plans we've already discussed in previous comments.

tampakrap:
> LGTM, with the follow-up plans we've already discussed in previous comments.

for the record, the follow-up plans are:
- replace the custom code with the actual reconciler to convert the Claim to XR https://github.com/crossplane/cli/pull/13#discussion_r3277384171
- introduce `--xrd` to automatically assign the proper kind https://github.com/crossplane/cli/pull/13#discussion_r3277333062

PR #24, opened 2026-05-21
Support pre-built function runtimes and per-language schema generation

### LINKED ISSUE #21: Support pre-built function runtime images in control plane projects

Per crossplane/crossplane#6840, Upbound is contributing the control plane project tooling to Crossplane. Part of this tooling is function building: taking function source code and producing OCI runtime images. The built-in builders work well for simple projects, but become limiting as projects grow in complexity.

For example, the Python builder detects `main.py`, tars the directory, and injects it into a base image. This is great for getting started, but it produces an on-disk layout that standard Python tools don't understand: type checkers can't resolve imports, pytest can't import function code, and there's no way to declare third-party dependencies. These limitations don't matter for simple projects, but they block standard development practices for complex ones.

More broadly, a complex Crossplane project may include components beyond composition functions — Go services, TypeScript frontends, or other tooling. The project already needs a build system (make, nix, CI pipelines) for these components. Having the Crossplane CLI also act as the function build tool means coordinating two build systems that may have conflicting opinions about project layout.

I'd like control plane projects to support pre-built function runtime images as inputs, alongside the existing source-directory-based building. Something like:

```yaml
spec:
  functions:
    # Source-based: the CLI builds the runtime image (today's behavior)
    - path: functions/function-a

    # Pre-built: the user provides the runtime image, the CLI packages it
    - image: build/function-b.oci.tar
```

For functions declared with `image:`, the CLI would skip the build step and use the provided OCI image as the function runtime. It would still add the Crossplane package metadata layer (crossplane.yaml) to produce a complete function package — the user provides the runtime, the CLI handles the packaging.

One consideration is multi-platform support. The built-in builders produce runtime images for multiple architectures (e.g. amd64, arm64). Pre-built images would need to support this too, either as a multi-arch OCI image index or as separate per-platform images.

This lets projects that outgrow the CLI's built-in builders use whatever build system they prefer (Docker, Nix, Bazel) while keeping the CLI's packaging, testing, and registry management. It also opens the door to function runtimes in languages the CLI doesn't have a built-in builder for.

The simple experience stays simple — `path:` functions are built by the CLI as today. Pre-built runtime images are the escape hatch for when you need full control over the build.

### LINKED ISSUE #29: Support generating schemas for specific languages

Both `crossplane dependency update-cache` and `crossplane project build` always generate schemas for all supported formats (Go, Python, KCL, and JSON Schema). In my project I only use the Python schemas — the rest are gitignored and unused. Generating all of them adds time to every invocation and leaves unused artifacts on disk.

This matters most when schema generation is part of an automated workflow (CI, a build system's `generate` step). Generating schemas I don't need slows down the loop and produces noise in the output.

A way to limit schema generation to specific formats. This could be:

A CLI flag on both commands:

```bash
crossplane dependency update-cache --language python
crossplane project build --language python
```

Or a project-level field in `crossplane-project.yaml`, since which schema languages a project uses is a property of the project rather than the invocation:

```yaml
spec:
  paths:
    schemaLanguages: [python]
```

Either way, the default (no flag / field omitted) should continue to generate all formats for backward compatibility.

negz:

This PR bundles two small improvements I wanted while taking the new Crossplane CLI for a test drive on a complex project. Each is in its own commit.

**1. Pre-built function runtime images** (fixes https://github.com/crossplane/cli/issues/21)

Crossplane projects today discover embedded functions by convention: every subdirectory of `paths.functions` is treated as a function, and the CLI auto-detects the language and builds the runtime image. This works well for simple projects but blocks projects that have outgrown the built-in builders or that need to coordinate function builds with an existing build system (make, nix, Bazel, CI pipelines).

This PR adds an optional `functions` list to `ProjectSpec`. When the list is present it disables auto-discovery and is the sole source of truth for which functions to build. Each entry uses a `source` discriminator (`Directory` or `Tarball`):

```yaml
spec:
  architectures: [amd64, arm64]
  functions:
    - source: Directory
      directory:
        name: function-a
    - source: Tarball
      tarball:
        name: function-b
        pathPrefix: build/function-b
```

Directory-source functions follow the existing build path. Tarball-source functions skip language detection and load one pre-built single-platform OCI image tarball per target architecture, following the naming convention `<pathPrefix>-<arch>.tar` or `<pathPrefix>-<arch>.tar.gz` (preferring the plain `.tar` when both exist). Per-architecture tarballs match what build tools naturally produce without bundling: `docker save`, Nix's `dockerTools.buildImage`, Bazel's `oci_tarball`, `ko build --tarball`, etc. all emit one single-platform tarball at a time. Packaging is inherently per-architecture too — each runtime image gets its own `crossplane.yaml` layer before they're tied together into a multi-arch package index — so the CLI would have to split a multi-arch input apart anyway. The gzipped variant is split into a separate commit; it's needed because Nix's image builders emit gzipped tarballs by default.

**2. Per-language schema generation** (fixes https://github.com/crossplane/cli/issues/29)

By default `crossplane project build` and `crossplane dependency update-cache` generate schemas for all four supported languages (Go, JSON, KCL, Python). For a project that only consumes one of them, every build generates language bindings the project never imports.

This commit adds an optional `schemas` block to `ProjectSpec`:

```yaml
spec:
  schemas:
    languages: [python]
```

When `languages` is set, schema generation is restricted to the listed languages, both for the project's own XRDs and for its declared dependencies. The filter flows through `project build`/`run` and `dependency update-cache`/`clean-cache`. The block is nested rather than flat to leave room for future schema-related knobs.

Reviewers may want to focus on `loadTarballRuntime` and `loadRuntimeImage` in `internal/project/build.go` (the new tarball loading path) and on `ProjectSchemas.Validate` in `apis/dev/v1alpha1/validate.go` together with `generator.Filter` in `internal/schemas/generator/interface.go` (the schema language filter). The language identifiers are now defined as constants in the API package and consumed by the generators directly, so the two can't drift.

I have:


internal/project/build.go
```diff
@@ -423,18 +446,24 @@ func (b *realBuilder) buildFunction(ctx context.Context, projectFS, fromFS afero
...
-		return nil, errors.Wrap(err, "failed to check for examples")
-	}
-	if examplesExist {
-		examplesParser = parser.NewFsBackend(fromFS,
-			parser.FsDir("/examples"),
-			parser.FsFilters(parser.SkipNotYAML()),
-		)
+	if fn.Source == devv1alpha1.FunctionSourceDirectory {
+		fnFS := afero.NewBasePathFs(functionsFS, fn.Directory.Name)
+		examplesExist, err := afero.IsDir(fnFS, "/examples")
```
adamwg: I'd mildly prefer to construct the examples directory path relative to the existing `functionsFS` filesystem rather than construct another `afero.BasePathFs` for this purpose. I believe this will be more efficient (since we'll `filepath.Join` once instead of afero doing it for every file operation), but more importantly it makes the ultimate path a little more obvious to readers.

`BasePathFs` has some quirks (especially when nested over/under other afero FS types) and I have, overall, been trying to reduce how much we rely on it. There's still some cleanup possible, as evidenced by the two `afero.Fs` arguments to `buildFunction`.

internal/project/build.go
```diff
@@ -448,37 +477,180 @@ func (b *realBuilder) buildFunction(ctx context.Context, projectFS, fromFS afero
...
-	return pkgImages, nil
+// loadRuntimeImage loads the runtime image for a single architecture, trying
+// the plain .tar file first and falling back to a gzipped .tar.gz file. It
+// returns the loaded image and the relative path (for error messages).
+func loadRuntimeImage(projectFS afero.Fs, prefix, arch string) (v1.Image, string, error) {
+	plain := fmt.Sprintf("%s-%s.tar", prefix, arch)
+	if exists, err := afero.Exists(projectFS, plain); err != nil {
+		return nil, plain, errors.Wrapf(err, "failed to stat runtime image %q", plain)
+	} else if exists {
+		img, err := tarball.ImageFromPath(resolveProjectPath(projectFS, plain), nil)
```
adamwg: We can simplify this (and the tests) by using `tarball.Image` with an opener that wraps the afero FS:

```go
opener := func() (io.ReadCloser, error) {
	return projectFS.Open(plain)
}
img, err := tarball.Image(opener, nil)
```

adamwg (commented):
Couple of small comments, but I like both of these features overall.

I wonder if `crossplane function generate` should refuse to generate functions in languages that aren't specified in `spec.languages`. Feels like it would be surprising to generate a function and find you don't have any schemas to use in it.

internal/project/build.go
negz: Done. I also added a separate commit that avoids passing around a separate filesystem for functions. So now we just pass around the project fs.

negz:
@adamwg I think I've addressed everything now.

adamwg (approved):
Thanks for the updates, this looks great.

PR #26, opened 2026-05-22
feat: add xpkg get-crds subcommand

### LINKED ISSUE #15: Command to download dependencies CRDs

When developing a composition, it's sometimes helpful to have the CRDs of the providers we use as dependencies handy.
Also, a flag to dump the JSON schema files instead of the CRDs would be really helpful.

A new command that does what `crossplane resource validate` (or `beta validate` used to do), that is, download the provider CRDs to a given folder.

For example, as `crossplane resource crds` or `crossplane xpkg crds`.

fernandezcuesta:

Add a `crossplane xpkg get-crds` subcommand to retrieve the CRD (and optionally, JSON schemas) of all dependencies (providers, functions).

Fixes #15

I have:

> Happy to do a PR for the docs, but I see the docs are not update with the last CLI changes?


adamwg (commented):
First off, thanks for the contribution! This is definitely a useful feature.

That said, there are a few high-level things I'd like to consider before reviewing this in detail:

1. The functionality here feels very close to `crossplane xpkg extract` in that it fetches xpkgs and extracts their contents (though it does very different things with that extracted contents). I wonder whether writing CRDs and JSONSchemas should become a feature of `extract` rather than a separate command.
2. We've also implemented OpenAPI -> JSONSchema conversion in `internal/schemas/generator`. It would be nice to re-use that code, which also handles a bunch of nuances to make the resulting JSONSchemas work super well with the YAML language server.
3. It's on my TODO list to change `crossplane validate` to use the shared infrastructure from `internal/xpkg`, which would also make it use the shared xpkg cache used by the `crossplane dependency` commands. It may actually make sense to use the existing `dependency.Manager` in validate, removing its custom `Manager` entirely. I don't think there's much harm in this PR introducing another usage of the `validate.Manager` (we can swap it out at the same time when we update `validate`), but I wonder if using the `dependency.Manager` would simplify things a bit here and also reduce the amount of consolidation needed in the future.

fernandezcuesta:
Hi @adamwg, thanks for taking a look.

Regarding your points:

1. Fully agree, moved it under `crossplane xpkg crd`. I'm still getting familiar with the command structure. Since the input is a `crossplane.yaml` file, I think it doesn't fit `xpkg extract`.
2. Refactor to reuse as much as possible. I made some slight changes to `oapiSchemaToJSONSchema` (apart from exposing it) to add $id and gvk info. Not fully needed but can be a useful addition IMO.
3. Agree this would be a better long term solution. Since you're already planning to move validate to the shared `internal/xpkg` infrastructure, I think the pragmatic path is to keep validate.Manager here for now and swap both at the same time. Happy to take on either approach though.

internal/schemas/generator/json.go
```diff
@@ -81,7 +82,11 @@ func (jsonGenerator) GenerateFromCRD(_ context.Context, fromFS afero.Fs, _ runne
 	return schemaFS, nil
 }
 
-func oapiSchemaToJSONSchema(s *spec.Schema) (*jsonschema.Schema, error) {
+// ToJSONSchema converts any JSON-compatible schema (e.g., *spec.Schema or
+// *extv1.JSONSchemaProps) to a *jsonschema.Schema with YAML language server
+// compatibility fixes applied. When group, version, and kind are non-empty, it
+// also sets $id and x-kubernetes-group-version-kind metadata.
+func ToJSONSchema(s any, group, version, kind string) (*jsonschema.Schema, error) {
```
adamwg: Nit: may be nice to take a `schema.GroupVersionKind` (from `k8s.io/apimachinery/pkg/runtime/schema`) rather than three strings, so that it's impossible to get the order wrong.

cmd/crossplane/xpkg/xpkg.go
```diff
@@ -29,6 +31,7 @@ type Cmd struct {
 	// Keep subcommands sorted alphabetically.
 	Batch   batchCmd   `cmd:"" help:"Batch build and push a family of provider packages."`
 	Build   buildCmd   `cmd:"" help:"Build a new package."`
+	CRD     crdCmd     `cmd:"" help:"Download CRDs from package dependencies."`
```
adamwg: Thanks for moving this into the `xpkg` command - definitely feels like the right place. I'm still not sure about the command name. Ideally, it would be a verb (or start with a verb) to match the general `crossplane <noun> <verb>` structure of the command tree, but I'm not sure what the right verb is. `extract` still feels like it could be right, but as you pointed out it doesn't quite fit with the current `extract` functionality.

A few ideas:

- `crossplane xpkg extract-crds`
- `crossplane xpkg extract-deps` (since it gets the dependencies of its argument)
- `crossplane xpkg get-crds`
- `crossplane xpkg dump-crds` (suggested by Claude)

Totally open to other suggestions.

cmd/crossplane/xpkg/crd.go
```diff
@@ -0,0 +1,253 @@
...
+	NoCache         bool   `help:"Disable caching entirely. Schemas are downloaded every time and not stored."`
+	OutputDir       string `default:"."                                                                                  help:"Directory where CRD or JSON Schema files will be written. Defaults to current directory." name:"output-dir"     short:"o"`
+	UpdateCache     bool   `default:"false"                                                                              help:"Update cached schemas by downloading the latest version that satisfies a constraint."`
+
+	fs afero.Fs
+}
+
+// Help prints out the help for the crd command.
+func (c *crdCmd) Help() string {
+	return `
```
adamwg: This text should be markdown formatted for integration into the auto-generated docs - see [Embedded Documentation](https://github.com/crossplane/cli/blob/main/CONTRIBUTING.md#embedded-documentation) in our contributing guidelines. (I think we introduced this after you originally wrote this code - sorry for the churn!)

adamwg (commented):
A few comments, but this looks good to me overall - thanks for the updates! Happy to leave it using the `cmd/validate` manager/cache for now and refactor to use the shared xpkg implementation in the future.

cmd/crossplane/xpkg/xpkg.go
fernandezcuesta: sure, renaming to `get-crds`

adamwg (approved):
LGTM, pending CI.

fernandezcuesta:
lint errors should be fine now :crossed_fingers:

fernandezcuesta:
@adamwg could you please approve to run the workfows? I expect this time to be compliant :crossed_fingers:

PR #49, opened 2026-05-27
Update CONTRIBUTING.md and add install instructions to README.md

adamwg:

Our original `CONTRIBUTING.md` doc just pointed to the core Crossplane
contributing documentation. Update it to also include some CLI-specific content.

While we're updating docs, add installation instructions to `README.md`.

I have: 


CONTRIBUTING.md
```diff
@@ -1 +1,114 @@
-See Crossplane's [contributing](https://github.com/crossplane/crossplane/tree/main/contributing).
+# Contributing to the Crossplane CLI
+
+Welcome to the Crossplane CLI, and thank you for your interest in contributing.
+
+Please start by reading the [Crossplane contributing document]. The Crossplane
```
tampakrap, line 5: I'd prefer this would be only a suggestion instead of a hard requirement. Newcomers might / should not know what is applicable and what is not from the other contributing docs, so I'd prefer to duplicate here the relevant parts (or to put specific links, but I guess it would be difficult to maintain, so I personally prefer the duplication)

CONTRIBUTING.md
```diff
@@ -1 +1,114 @@
...
+go.
+
+## kong
+
+We use [kong] as our CLI framework. Each command is defined as a struct, in
+which fields become subcommands, positional arguments, or flags. Kong's [struct
+tags] can be used to control many behaviors, including validation,
+auto-completion, and documentation.
+
+## Command Maturity Levels
```
tampakrap, line 29: We should also mention that commands without maturity level inherit it from their parent

CONTRIBUTING.md
```diff
@@ -1 +1,114 @@
...
+  `generate-docs` command to ensure the documentation page is properly
+  structured.
+- Use inline code and code blocks for commands, flags, arguments, etc. This
+  looks good when rendered, and avoids spelling and grammar false positives.
+
+If in doubt about whether your help will look good, you can easily check your
+work by cloning the [crossplane/docs](https://github.com/crossplane/docs) repo,
+running `crossplane generate-docs -o content/master/cli/command-reference.md` to
+update the CLI reference section, and then running `hugo serve` to preview it
+locally.
```
tampakrap: maybe to convert that to a list of commands? So that it is more clear on where to run each command

tampakrap (commented):
in general it looks good, I left some minor comments, thanks for taking the time to write this!

Another comment (that can be addressed in a follow-up PR), we should add instructions for development, eg how to run tests and build/install a local version.

CONTRIBUTING.md
adamwg: Good idea, added a paragraph about this.

CONTRIBUTING.md
adamwg: Good idea, done.

CONTRIBUTING.md
adamwg: I think this makes sense. I worry a bit about coding style guidelines in particular diverging, but the coding style in Crossplane is pretty settled and evolves slowly, so I think the risk is low.

Can we handle this as a follow-up? I think we can copy over most of what we need from c/c, but organizing the content nicely will take some thought.

adamwg:
Rebased to (hopefully) make the `validate-docs` check pass.

CONTRIBUTING.md
```diff
@@ -1 +1,129 @@
...
+- Alpha: Experimental commands that may change significantly or be removed at
+  any time.
+
+Alpha and beta commands are marked as such using a custom kong struct tag,
+`maturity:"alpha"` or `maturity:"beta"`. These struct tags control two things:
+
+1. Alpha commands are hidden from the help output by default. Users must enable
+   showing alpha commands in their CLI configuration file.
+2. Alpha and beta commands automatically have notes added to their help text
+   indicating the maturity level and what it means.
```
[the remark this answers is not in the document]
adamwg: The beta level is enabled by default.

CONTRIBUTING.md
tampakrap: I believe coding style (and maybe more parts that are equally stable) could possibly be just a link, but an explicit reference so that it's clear that they are relevant to the CLI as well. In any case, I'm fine handling that in a followup PR so we can proceed with this one.

tampakrap (approved):
thanks for addressing my comments! LGTM, the rest can be handled in followup PRs

PR #61, opened 2026-06-02
feat(xr): Introduce `xr patch --xrd`

tampakrap:

New subcommand that does the same as `render --xrd` but without running render. Useful when we want to patch our XR with its default values from the XRD.

It is using the same code as `render --xrd`, but that codeis moved under `xr/xrd.go` and `render --xrd` imports/uses it.

I have: 


cmd/crossplane/xr/xrd.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return DefaultValues(xr.UnstructuredContent(), xr.GetAPIVersion(), *crd)
+}
+
+// DefaultValues sets default values on the XR based on the CRD schema.
+//
+// Callers starting from an XRD should prefer ApplyXRDDefaults; this is the
+// lower-level routine for callers that already have a CRD in hand.
+func DefaultValues(xr map[string]any, apiVersion string, crd extv1.CustomResourceDefinition) error {
```
tampakrap: I kept the same name and code that it was there. But maybe we should rename this function to `ApplyCRDDefaults` to match the other? If we do that though, we might break someone's setup that might be using it.

adamwg: We'll already have broken them by moving it from `cmd/crossplane/render` to `cmd/crossplane/xrd`, so I'd be in favor of renaming as well.

Might also be a good time to move this code out of `cmd/` since multiple commands rely on it. Maybe a new exported package like `pkg/xr` or `pkg/composite`.

cmd/crossplane/xr/patch.go
```diff
@@ -0,0 +1,104 @@
...
+
+type patchCmd struct {
+	// Arguments.
+	InputFile string `arg:"" default:"-" help:"The XR YAML file to patch, or '-' for stdin." optional:"" predictor:"file" type:"path"`
+
+	// Output flags.
+	OutputFile string `help:"The file to write the patched XR YAML to. Defaults to stdout." placeholder:"PATH" predictor:"file" short:"o" type:"path"`
+
+	// Patching flags.
+	XRD string `help:"A YAML file specifying the CompositeResourceDefinition (XRD) whose schema defaults are applied to the XR." name:"xrd" placeholder:"PATH" predictor:"file" type:"path"`
```
adamwg: Can we give this a more descriptive name like `--defaults-from-xrd` or just `--xrd-defaults`? It's not really clear from the flag name `--xrd` that it indicates an action.

tampakrap: I chose `xr patch --xrd` to be the same as `render --xrd`, which is not descriptive either, but people know it very well. In theory, in the future passing the XRD could mean additional things apart from applying the defaults. Furthermore, I don't expect this command to be heavily used, I believe it will be used only for debugging purposes. The libraries behind this command though are going to be used in the upcoming `test` subcommand, which will have `patch.xrd` as defined in our design doc.

That said, I'm fine using a more descriptive name to the flag, I just wanted to give that background before we take a decision though.

cmd/crossplane/xr/xrd.go
tampakrap: good idea, what is your preference?

tampakrap: renamed and both moved to `pkg/xr`. I chose that to avoid potential confusion/conflicts with `crossplane-runtime/pkg/resource/unstructured/composite`

adamwg (approved):
lgtm.

cmd/crossplane/xr/patch.go
adamwg: True, makes sense to match `render`'s flag and leave it open to do more in the future. I'm fine leaving this as-is.

PR #65, opened 2026-06-03
fix(render): do not overwrite function docker network if set, start crossplane-container in same network

### LINKED ISSUE #75: bug(render/v2.3.0): function docker network is overwritten and crossplane container always start in temporary network
### What happened?

Since v2.3.0, crossplane render uses crossplane's engine to render the compositions. By default this is done by starting up crossplane as its own container in a temporary network, and overwriting the docker-network annotation on all functions to join it.

This new default behavior has two problems we have encountered:

1) This doesnt support starting/stopping functions yourself so that you can run multiple renders without starting/stopping the functions on each run.
Since the temporary crossplane-render network and crossplane-container is created on each render-run you'd need to connect the containers to this new network, for every render.

2) Running crossplane render in a devcontainer / docker-in-docker / podman-in-podman
If you run crossplane-render inside a container, you need to ensure that containers started by crossplane-render run in the same network as the container you are in (or in a connected network).
Previously you could control this with the docker-network annotation, either with -a or as annotation in the functions-file, but since the crossplane-container always starts up in a temporary network, there's no connectivity to the crossplane-container. If function-containers already run, i'm unsure if the annotation overwriting has any effect.
Circumventing this would require a bit of hacking.

### How can we reproduce it?

Run crossplane-render in ci (DinD) or devcontainer.

* Crossplane CLI version: v2.3.1

nkzk:

Closes https://github.com/crossplane/cli/issues/75

Fixes:

- Do not overwrite the docker-network annotation in functions if it has already been set
- If the docker-network annotation is passed to the `FunctionAnnotations` flag, run crossplane-container in it.

I have: 

- Added `backport release-x.y` labels to auto-backport this PR.

### DIFF
--- a/cmd/crossplane/render/engine.go
+++ b/cmd/crossplane/render/engine.go
@@ -66,9 +66,7 @@ type EngineFlags struct {
 
 // NewEngineFromFlags creates an Engine from the flag configuration. If a binary
 // path is set, it returns a local engine. Otherwise it returns a Docker engine
-// using the resolved image reference. The network parameter sets the Docker
-// network the render container should join; it is derived from function
-// annotations (AnnotationKeyRuntimeDockerNetwork) by the caller.
+// using the resolved image reference.
 func NewEngineFromFlags(f *EngineFlags, log logging.Logger) Engine {
 	if f.CrossplaneBinary != "" {
 		return &localRenderEngine{BinaryPath: f.CrossplaneBinary}
--- a/cmd/crossplane/render/engine.go
+++ b/cmd/crossplane/render/engine.go
@@ -58,9 +58,9 @@ type Engine interface {
 // EngineFlags contains flags for configuring the render engine. It is embedded
 // by render command structs to provide shared engine configuration.
 type EngineFlags struct {
-	CrossplaneVersion       string `help:"Version of the Crossplane image to use for rendering. Defaults to the latest stable version." placeholder:"VERSION" xor:"crossplane-selector"`
-	CrossplaneImage         string `help:"Override the full Crossplane Docker image reference for rendering."                           placeholder:"IMAGE"   xor:"crossplane-selector"`
-	CrossplaneBinary        string `help:"Path to a local crossplane binary to use instead of Docker."                                  placeholder:"PATH"    type:"existingfile"       xor:"crossplane-selector,crossplane-docker"`
+	CrossplaneVersion       string `help:"Version of the Crossplane image to use for rendering. Defaults to the latest stable version." placeholder:"VERSION"   xor:"crossplane-selector"`
+	CrossplaneImage         string `help:"Override the full Crossplane Docker image reference for rendering."                           placeholder:"IMAGE"     xor:"crossplane-selector"`
+	CrossplaneBinary        string `help:"Path to a local crossplane binary to use instead of Docker."                                  placeholder:"PATH"      type:"existingfile"       xor:"crossplane-selector,crossplane-docker"`
 	CrossplaneDockerNetwork string `help:"The docker network to start the crossplane container in"                                      xor:"crossplane-docker"`
 }
 
--- a/cmd/crossplane/render/engine.go
+++ b/cmd/crossplane/render/engine.go
@@ -61,12 +61,13 @@ type EngineFlags struct {
 	CrossplaneVersion string `help:"Version of the Crossplane image to use for rendering. Defaults to the latest stable version." placeholder:"VERSION" xor:"crossplane-selector"`
 	CrossplaneImage   string `help:"Override the full Crossplane Docker image reference for rendering."                           placeholder:"IMAGE"   xor:"crossplane-selector"`
 	CrossplaneBinary  string `help:"Path to a local crossplane binary to use instead of Docker."                                  placeholder:"PATH"    type:"existingfile"       xor:"crossplane-selector"`
-	Network           string `help:"The network containers should connect to"`
 }
 
 // NewEngineFromFlags creates an Engine from the flag configuration. If a binary
 // path is set, it returns a local engine. Otherwise it returns a Docker engine
-// using the resolved image reference.
+// using the resolved image reference. The network parameter sets the Docker
+// network the render container should join; it is derived from function
+// annotations (AnnotationKeyRuntimeDockerNetwork) by the caller.
 func NewEngineFromFlags(f *EngineFlags, network string, log logging.Logger) Engine {
 	if f.CrossplaneBinary != "" {
 		return &localRenderEngine{BinaryPath: f.CrossplaneBinary}
--- a/cmd/crossplane/render/engine.go
+++ b/cmd/crossplane/render/engine.go
@@ -58,9 +58,10 @@ type Engine interface {
 // EngineFlags contains flags for configuring the render engine. It is embedded
 // by render command structs to provide shared engine configuration.
 type EngineFlags struct {
-	CrossplaneVersion string `help:"Version of the Crossplane image to use for rendering. Defaults to the latest stable version." placeholder:"VERSION" xor:"crossplane-selector"`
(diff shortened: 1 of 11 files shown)


adamwg:
Thanks for the PR, @nkzk! Would you mind creating an issue for this as well, for discoverability and tracking? I haven't reviewed in detail yet, but the described fixes sound reasonable.

nkzk:
I see that there are already tests for passing function annotations to the engine. I had copilot help me create unit tests for injectNetworkAnnotations. Also ran flake check.

nkzk:
Hmm, i got it to work in a devcontainer with this fix, but the current implemention has some issues in CI. But i think this can be solved on the user-side.

One of our earliest approaches was to start up functions as service-containers before running multiple/different renders, and it worked because gitlab/github connects the job-container to the bridge-network used by service-containers.

But since crossplane render will start up crossplane in another temporary bridge network, it doesnt seem that this will continue to work. However, my theory is that the user can specify the docker-network in their CI-provider (gitlab/github), and then specify the the docker-network flag in the crossplane render command with the fix in this branch to solve this.

We have another workflow which uses rootless DinD/PinP, but kind of the same issue there.

I'll do some more testing soon.

But let me know if something i say sounds off :D

nkzk:
Think this PR is ready for review, i did some more testing in CI and did not completely figure it out yet, but i think its just an issue of configuring the docker-network in CI and setting that value as the function-docker-network flag in the render-command.

A quality of life improvement for us would be if we can spin up the crossplane container ourselves and make render use it. If we could configure the crossplane-containerthe same way as functions, with the development annotation to manage the container lifecycle ourselves, it would just simplify this alot for us.

But maybe its out of scope for this PR, i'm not sure whats the best way to implement this would be. But open to work on it if someone has some ideas.

adamwg:
> A quality of life improvement for us would be if we can spin up the crossplane container ourselves and make render use it. If we could configure the crossplane-containerthe same way as functions, with the development annotation to manage the container lifecycle ourselves, it would just simplify this alot for us.
> 
> But maybe its out of scope for this PR, i'm not sure whats the best way to implement this would be. But open to work on it if someone has some ideas.

@nkzk Good thought - I can see how this would be useful. It's a little tricky, since the crossplane container in `render` doesn't actually run a server, it's just a one-off command (`crossplane internal render ...`).

For your use-case, would it be easier to download a crossplane binary and use the `--crossplane-binary` render flag? In that mode, the functions need to be accessible to the host (like with the old `crossplane render`), but there's no assumptions about inter-container networking.

cmd/crossplane/render/engine_docker.go
```diff
@@ -83,19 +82,28 @@ func (e *dockerRenderEngine) CheckContextSupport() error {
 // containers also join it. The returned cleanup function removes the
 // network.
 func (e *dockerRenderEngine) Setup(ctx context.Context, fns []pkgv1.Function) (func(), error) {
-	networkID, networkName, err := createRenderNetwork(ctx)
-	if err != nil {
-		return func() {}, errors.Wrap(err, "cannot create Docker network for rendering")
-	}
+	var networkID, networkName string
+
+	if e.network == "" {
```
adamwg: I think this will also fix #96 :-).

Style nit: prefer to return early and keep the larger logic outdented, so:

```go
if e.network != "" {
	return func() {}, nil
}
// Network creation logic goes here.
```

cmd/crossplane/render/engine.go
```diff
@@ -65,13 +65,15 @@ type EngineFlags struct {
 
 // NewEngineFromFlags creates an Engine from the flag configuration. If a binary
 // path is set, it returns a local engine. Otherwise it returns a Docker engine
-// using the resolved image reference.
-func NewEngineFromFlags(f *EngineFlags, log logging.Logger) Engine {
+// using the resolved image reference. The network parameter sets the Docker
+// network the render container should join; it is derived from function
+// annotations (AnnotationKeyRuntimeDockerNetwork) by the caller.
+func NewEngineFromFlags(f *EngineFlags, network string, log logging.Logger) Engine {
```
adamwg: I'd prefer not to introduce a new string argument here. Two possible options:

1. Add `DockerNetwork` to `EngineFlags`. This would let users specify `--docker-network` on the command-line, which seems like it could be useful, but `Run` can still infer it from function annotations by default.
2. Introduce functional arguments (`EngineWithNework(...)`) here, so that we have a place to put all the possible optional things. `WithLogger` could become one of these, simplifying the signature futher.

I think I lean toward option 1 since it has the side-effect of letting users specify the docker network name as a flag.

cmd/crossplane/render/xr/cmd.go
```diff
@@ -224,7 +224,13 @@ func (c *Cmd) Run(k *kong.Context, log logging.Logger, sp terminal.SpinnerPrinte
 		}
 	}
 
-	engine := c.newEngine(&c.EngineFlags, log)
+	var network string
+	annotations := render.NewAnnotationsFromStrings(c.FunctionAnnotations)
```
adamwg: Should we also consider network annotations on the functions in the functions file? `c.FunctionAnnotations` would override those, but it seems like it would be nice to have render work by default when all functions are already annotated with the same network name.

adamwg (commented):
Thanks for filing an issue for this, and for the fix. A few comments inline, but the overall approach looks good to me.

cmd/crossplane/render/op/cmd.go
```diff
@@ -167,6 +167,24 @@ func (c *Cmd) Run(k *kong.Context, log logging.Logger, sp terminal.SpinnerPrinte
...
+			}
+		}
+
+		// Overwrite with docker-network annotation from function-annotations cli flag if set
+		if len(c.FunctionAnnotations) > 0 {
+			annotations := render.NewAnnotationsFromStrings(c.FunctionAnnotations)
+			if value, ok := annotations[render.AnnotationKeyRuntimeDockerNetwork]; ok {
+				c.EngineFlags.CrossplaneDockerNetwork = value
+			}
+		}
```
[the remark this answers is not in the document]
nkzk: We only want the override behavior if c.EngineFlags.CrossplaneDockerNetwork is not set, so i dont think this is a mistake. It only overrides the network that was set based on the first docker-network function-annotation in the provided functions.

jcogilvie:
I like this change and I want to see it land, but you've checked off the PR checklist box about adding tests and I don't see any.  Have I missed something?

adamwg (approved):
LGTM. Per @jcogilvie's comment, this could probably use a bit of test coverage (may be easiest to factor out the network defaulting logic from both render commands and test it separately).

nkzk:
yeah my bad, had some tests but dissapeared in a revert+refactor.  i can look into it tomorrow!

cmd/crossplane/render/network.go
```diff
@@ -28,6 +29,22 @@ import (
 	"github.com/crossplane/cli/v2/internal/docker"
 )
 
+// SetDefaultCrossplaneDockerNetwork defaults the Crossplane render engine's
+// Docker network to the first function runtime Docker network annotation.
+// If network is already set, return with no changes
```
nkzk:
```suggestion
// If network is already set, return with no changes.
```

nkzk:
Added tests with alot of help from copilot, but the tests look reasonable to me. Let me know if you have any more feedback, thanks! :)

cmd/crossplane/render/network.go
```diff
@@ -25,9 +25,27 @@ import (
...
+// If network is already set, return with no changes.
+func (f *EngineFlags) SetDefaultCrossplaneDockerNetwork(fns []pkgv1.Function) {
+	if f.CrossplaneDockerNetwork != "" {
+		return
+	}
+
+	for _, fn := range fns {
+		if value, ok := fn.Annotations[AnnotationKeyRuntimeDockerNetwork]; ok {
+			f.CrossplaneDockerNetwork = value
+			return
```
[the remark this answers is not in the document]
nkzk: lgtm, didnt know empty strings returned ok, but it does https://go.dev/play/p/IjtUe8Y0XYd

PR #66, opened 2026-06-03
feat(validate): structured results and --output json|yaml flag

### LINKED ISSUE #62: Add machine-readable output for `validate`

I would like to use `crossplane validate` in CI scripting and downstream code.  However, right now I need to parse text from stdout in order to do so.

Add flags for yaml/json output from validate.  

Bonus:  allow downstream packages to import and call validate from golang w/o going through any parsing at all.  Expose an interface that returns meaningful results.

jcogilvie:

Adds machine-readable output to `crossplane resource validate` and exposes the validation logic as a programmatic Go API.

The user-facing change is a `-o/--output text|json|yaml` flag. The default text output is byte-identical to the historical format (same `[✓]`/`[x]`/`[!]` markers, same line order, same summary line); JSON and YAML emit the full structured result, including per-resource status and field-level error details — type, field path, message, and bad value.

The structural changes that enable it:

- **Processing API** at `cmd/crossplane/pkg/validate/` — `SchemaValidate(ctx, resources, crds) (*ValidationResult, error)` returns structured data with no I/O. Programmatic consumers (the motivating example: [crossplane-diff](https://github.com/crossplane-contrib/crossplane-diff)) call it directly and inspect the result, replacing the previous parse-stdout pattern.
- **Rendering API** at `cmd/crossplane/pkg/validate/render/` — `RendererFor(OutputFormat) (Renderer, error)` returns the format-specific renderer. Lives in a sibling subpackage so consumers depending only on validation data don't pull in YAML/JSON encoding deps.
- **CLI wiring** — the `--output` flag is decoded directly into a typed `render.Renderer` via Kong's `MapperValue` interface, so `Cmd.Run` works against the dependency rather than threading a format string through validation logic. Exit codes and `--skip-success-results` semantics are unchanged.

**Breaking change for direct Go callers**: the previous `validate.SchemaValidation(ctx, resources, crds, errorOnMissingSchemas, skipSuccessLogs, w)` entry point is removed. Callers migrate to `pkgvalidate.SchemaValidate` + a renderer from `render.RendererFor` + `pkgvalidate.ResultError`. The CLI surface (text bytes, exit codes, flag semantics) is preserved.

Sample JSON output:

```json
{
  "summary": { "total": 2, "valid": 1, "invalid": 1, "missingSchemas": 0 },
  "resources": [
    { "apiVersion": "smoke.org/v1alpha1", "kind": "Test", "name": "ok-instance", "status": "valid" },
    {
      "apiVersion": "smoke.org/v1alpha1", "kind": "Test", "name": "bad-instance", "status": "invalid",
      "errors": [
        { "type": "schema", "field": "spec.replicas",
          "message": "spec.replicas: Invalid value: \"string\": spec.replicas in body must be of type integer: \"string\"",
          "value": "string" }
      ]
    }
  ]
}
```

**Tests**

- `pkg/validate/validate_test.go`: `TestSchemaValidate` — 9 table-driven cases covering Valid, InvalidSchema, InvalidCEL, MissingSchema, UnknownField, DefaultingFailureOnly, DefaultingFailureWithSchemaError, Empty, and MixedOrder. `TestResultError` — 5 cases covering Clean, InvalidPresent, MissingIgnored, MissingWithFlag, InvalidAndMissing.
- `pkg/validate/{apply_defaults,unknown_fields}_test.go`: `TestApplyDefaults` and `TestValidateUnknownFields` moved here alongside the helpers they test.
- `pkg/validate/render/render_test.go`: `TestRendererFor_Text` (per-line substring assertions, not full-string golden), `TestRendererFor_JSON`, `TestRendererFor_YAML` (round-trip), `TestRendererFor_FormatBoundary` (text/json/yaml/empty/unknown).
- `validate/validate_integration_test.go`: a Kong-driven end-to-end test (`TestRun`) that runs `crossplane resource validate` against fixture YAML files. The pre-populated `testdata/cache/` directory keeps the run offline. 8 subtests cover text/json/yaml × valid/invalid/missing × `--skip-success-results` and `--error-on-missing-schemas`. `TestParse_RejectsUnknownOutputFormat` covers parse-time rejection of unknown formats by `rendererFlag.Decode`.
- The previously existing `TestValidateResources` is removed — it tested the now-deleted `SchemaValidation` wrapper, and `TestSchemaValidate` covers the same logic with stronger structural assertions. `TestConvertToCRDs` and the manager / cache / image tests pass unchanged.

**Reviewer attention**

- Defaulting-failure semantics intentionally match the historical CLI behavior. A defaulting-only failure renders as `[!]` (warning) and counts as a success in the summary; a defaulting failure combined with a schema error produces both lines and counts as a failure. `DefaultingFailureWithSchemaError` is the regression guard that schema validation still runs after a defaulting failure.
- `validate_integration_test.go` ships a stand-in `crossplane@v0.0.0-test/package.yaml` under `testdata/cache/` so the test runs offline through the real `Manager`/`LocalCache` code paths. Alternative would have been adding a `WithFetcher` option to `Manager` for test injection — chose the cache fixture as the smaller change that exercises more production code.
- Help text (`cmd/crossplane/validate/help/validate.md`) was updated with a `--output json | jq` example.

Fixes #62

I have:

- [x] Linked a PR or a [docs tracking issue] to [document this change]. — [crossplane/docs#1101](https://github.com/crossplane/docs/issues/1101)
- [ ] ~~Added `backport release-x.y` labels to auto-backport this PR.~~


cmd/crossplane/validate/cmd.go
```diff
@@ -32,26 +32,63 @@ import (
...
 
 	_ "embed"
 )
 
 //go:embed help/validate.md
 var helpDetail string
 
+// errWriteOutput is the error message wrapped around I/O failures when the
+// validate command writes to its output writer.
+const errWriteOutput = "cannot write output"
```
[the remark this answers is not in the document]
jcogilvie: Skipping this one — `errWriteOutput` is used in [`manager.go:315`](https://github.com/crossplane/cli/blob/11df1ed/cmd/crossplane/validate/manager.go#L315) (the "schemas does not exist, downloading: ..." Fprintln wrap), so removing it would break the build. The constant lives in `cmd.go` because the prior monolithic `validate.go` was where it was originally declared; `cmd.go` is now the only top-level non-test file in the package, hence the placement. Happy to move it to a dedicated `errors.go` if you'd prefer it not appear adjacent to a Cmd that doesn't use it.

cmd/crossplane/pkg/validate/render/render.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/crossplane/crossplane-runtime/v2/pkg/errors"
+
+	pkgvalidate "github.com/crossplane/cli/v2/cmd/crossplane/pkg/validate"
+)
+
+const (
+	errCannotMarshalJSON = "cannot marshal validation result as JSON"
+	errCannotMarshalYAML = "cannot marshal validation result as YAML"
+	errCannotWriteOutput = "cannot write validation result"
+	errUnknownFormat     = "unknown output format"
```
tampakrap: https://github.com/crossplane/crossplane/blob/main/contributing/README.md#wrap-errors

pkg/validate/output/render.go
```diff
@@ -0,0 +1,177 @@
...
+// a bare string so that call sites can use the symbolic constants below
+// (OutputFormatText, OutputFormatJSON, OutputFormatYAML) instead of
+// embedding raw "text"/"json"/"yaml" literals, and so the compiler
+// catches accidental cross-wiring of unrelated string flags.
+type OutputFormat string
+
+// OutputFormat values.
+const (
+	// OutputFormatText renders results in human-readable text format with
+	// [x], [!], [✓] markers.
```
tampakrap, line 66: how about using constants also for the markers?

cmd/crossplane/pkg/validate/render/render.go
jcogilvie: Read the link as flagging the `failed to` prefix; I had two of those in [`validate.go`'s `applyDefaults`](https://github.com/crossplane/cli/blob/2b3f641/cmd/crossplane/pkg/validate/validate.go#L300-L305). Both swapped to `cannot ...` to match the rest of the package's wrap style. Pushed in 2b3f641.

(🤖)

pkg/validate/output/render.go
jcogilvie: Done — added `markerSuccess`/`markerWarning`/`markerError` constants and used them across `Render` and `textErrorLine`. Constants concatenate at compile time so the format strings still read top-to-bottom. Pushed in 2b3f641.

(🤖)

cmd/crossplane/pkg/validate/render/render.go
jcogilvie: thanks, fixed!

tampakrap: apologies for the lack of context, I meant that the coding style is not to use constants any more and instead use inline errors

pkg/validate/output/render.go
tampakrap: The markerWarning is not really a warning, it is an operational error. I'd suggest to use the naming markerError for [!] (test/assertion operational error, which can cause a failure when the appropriate flag is passed) and markerFail for [x] (test/assertion failure)

tampakrap (commented):
I left a couple of comments, apart from that I don't see any other issue. I plan also to give it a spin today to make sure it works fine

pkg/validate/output/render.go
tampakrap: on a second thought, let's keep markerWarning because even if it is an operational error, in the end we treat it as a warning. I'd still rename markerError to markerFail though

cmd/crossplane/pkg/validate/render/render.go
jcogilvie: done in d192b93.

pkg/validate/output/render.go
jcogilvie: renamed per suggestion; updated comment block to clarify (`[!]` = operational issue, `[x]` = assertion failure). d192b93.

tampakrap (approved):
LGTM thank you! I tested it locally, passing all the three options of the new `--output` flag and got the expected results. One nitpick would be not to use `render` for the package name to avoid confusion with the popular subcommand, but that's minor so I'll approve and let you decide if you want to keep that name or change it

tampakrap (commented):
one more comment: how about moving `cmd/crossplane/pkg/validate` to topdir `pkg/validate`?

jcogilvie:
@tampakrap done in 275bda9 — moved `cmd/crossplane/pkg/validate` → top-level `pkg/validate`, and renamed the `render` subpackage to `output` (with `OutputFormat` → `Format`, `OutputFormatText` → `FormatText`, etc., since the type now lives inside `output`). `nix flake check` still passes. Thanks for the review!

PR #74, opened 2026-06-04
render: Sync render.proto with c/c and wire XRD through

jcogilvie:

Two commits:

1. **`proto: Sync render.proto with crossplane/crossplane main`** — adds `composite_resource_definition` (field 8) to `CompositeInput`, matching the upstream proto, and regenerates `render.pb.go` via `./nix.sh run .#generate` (Nix-pinned `protoc-gen-go` v1.36.10). The cli's `option go_package` intentionally still points at `github.com/crossplane/cli/v2/proto/render/v1alpha1`; after that one line the proto file is byte-identical to upstream `main`.

2. **`render: wire CompositionInputs.XRD through to the proto`** — adds an `XRD *unstructured.Unstructured` field to `CompositionInputs` and threads it through `BuildCompositeRequest` to the proto's `CompositeResourceDefinition`. When the caller leaves `XRD` nil, behavior is unchanged: the field stays unset and the render binary keeps its `Schema=Modern` fallback. This mirrors the Go-level change in `crossplane/crossplane/cmd/crossplane/render` so the cli's `CompositionInputs` stays in sync with what the binary now consumes.

Picking up an XRD from disk / the `--extra-resources` flag and populating `CompositionInputs.XRD` at the call sites (e.g. `cmd/crossplane/render/xr/cmd.go`) is intentionally out of scope here — that's a separate UX-shaped change. After this PR the plumbing is in place; the field is available to any caller that wants to use it.

Upstream proto changes:
- crossplane/crossplane@12f3f1c (adds the field; `composite_resource_definitions` plural at the time)
- crossplane/crossplane@babb251 (review feedback: collapses to singular `composite_resource_definition`; this PR matches the post-feedback shape on `main`)

How this is covered by tests: regen is mechanical, and the wiring commit is pure plumbing — leaving `CompositionInputs.XRD` nil produces the same proto request as before, so the existing `cmd/crossplane/render/...` tests cover the no-XRD path. `./nix.sh flake check` passes.

Fixes #

I have:

- [ ] ~~Added or updated unit tests.~~ Regen + nil-default plumbing; the existing render tests cover the unchanged no-XRD path. Happy to add a test that asserts the XRD-set path produces the populated proto field if reviewers want it.
- [ ] ~~Linked a PR or a [docs tracking issue] to [document this change].~~ No user-facing change yet — this PR adds the proto field and the `CompositionInputs.XRD` plumbing; there's no command-surface change to document.
- [ ] ~~Added `backport release-x.y` labels to auto-backport this PR.~~ Not a backport candidate; happy to add a label if maintainers want it on a release branch.

### DIFF
--- a/cmd/crossplane/render/convert.go
+++ b/cmd/crossplane/render/convert.go
@@ -77,17 +77,26 @@ func BuildCompositeRequest(in CompositionInputs) (*renderv1alpha1.RenderRequest,
 		return nil, errors.Wrap(err, "cannot convert required schemas to protobuf")
 	}
 
+	var xrdStruct *structpb.Struct
+	if in.XRD != nil {
+		xrdStruct, err = resource.AsStruct(in.XRD)
+		if err != nil {
+			return nil, errors.Wrap(err, "cannot convert XRD to protobuf")
+		}
+	}
+
 	return &renderv1alpha1.RenderRequest{
 		Meta: &renderv1alpha1.RequestMeta{},
 		Input: &renderv1alpha1.RenderRequest_Composite{
 			Composite: &renderv1alpha1.CompositeInput{
-				CompositeResource: xrStruct,
-				Composition:       compStruct,
-				Functions:         fnInputs,
-				ObservedResources: observedStructs,
-				RequiredResources: requiredStructs,
-				RequiredSchemas:   schemaStructs,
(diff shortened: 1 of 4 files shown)


adamwg (approved):
lgtm, thanks for taking care of this.

PR #82, opened 2026-06-05
render: Replace condition timestamps and sort resources to stabilize output

adamwg:
### Description of your changes

The old render command used a static timestamp for status condition `lastTransitionTime`s so that output was stable across runs with the same inputs. The new render engine calls the real reconciler, and therefore produces real timestamps.

Restore stability by replacing the timestamps in any status conditions returned by the render engine before outputting them. We use the same static timestamp that was present in the old CLI for consistency with its output.

Fixes #50

The old CLI's `render` command sorted composed resources by the `composition-resource-name` annotation to ensure output consistency across runs with the same inputs. The real reconciler code used by the new render engine does not sort the results, so we need to sort them client-side to ensure stability.

Fixes #53

I have: 

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `./nix.sh flake check` to ensure this PR is ready for review.
- [x] Added or updated unit tests.
~- [ ] Linked a PR or a [docs tracking issue] to [document this change].~
- [x] Added `backport release-x.y` labels to auto-backport this PR.

[contribution process]: https://github.com/crossplane/crossplane/tree/main/contributing
[docs tracking issue]: https://github.com/crossplane/docs/issues/new
[document this change]: https://docs.crossplane.io/contribute/contribute

PR #83, opened 2026-06-05
render: Clean up unused code and duplicate consts

adamwg:

The coderabbit bot noted on #82 that we're duplicating the
`xcrd.AnnotationKeyCompositionResourceName` const from crossplane-runtime. The
first commit in this PR removes this duplication.

While digging into the duplicate const, I noticed that we have a bunch of
duplicated consts, which led me to a couple of functions that are exported but
unused since we switched to `crossplane internal render`. These functions just
duplicate functionality from the real composite controller, so they shouldn't be
necessary for any external consumers of the render code.

I have: 


tampakrap (approved):
🧹

PR #91, opened 2026-06-08
feat(render): capture stderr and handle pipeline-fatal exit code in both engines

jcogilvie:

The `crossplane internal render` engines (`localRenderEngine` in `engine_local.go` and `dockerRenderEngine` in `engine_docker.go`) currently drop two kinds of information that callers and tooling need:

1. **stderr is not surfaced in returned errors.**
   - `engine_local.go` sets `cmd.Stderr = os.Stderr`, so any FATAL message printed by the binary is visible on the terminal but never appears in the returned Go error.
   - `engine_docker.go` discards `RunContainer`'s stderr return value at the engine layer (`stdout, _, err := docker.RunContainer(...)`); only on container exit-non-zero does stderr appear, folded into an opaque `fmt.Errorf` string.

2. **The pipeline-fatal exit-code-3 contract is not honoured.**
   crossplane/crossplane#7455 made `crossplane internal render` exit with code 3 (instead of 1) when a pipeline step returns `SEVERITY_FATAL`, with stdout populated by a partial `RenderResponse` so callers can recover `output.RequiredResources` and iterate. Both engines currently wrap the resulting `*exec.ExitError` / opaque container error and discard stdout, dropping the partial response on the floor.

#### Changes

- `cmd/crossplane/render/render.go`: add `ExitCodePipelineFatal = 3`, with a doc comment pointing at crossplane/crossplane#7455 so the contract is discoverable.
- `internal/docker/docker.go`: introduce `*ContainerExitError` (carrying `ExitCode` and captured stderr). `RunContainer` now returns it on non-zero exit instead of `fmt.Errorf`, so callers can `errors.As` against it and branch on the exit code.
- `cmd/crossplane/render/engine_local.go`: capture stderr into a `bytes.Buffer` instead of piping to `os.Stderr`. On exit code 3 with non-empty stdout, parse the partial `*RenderResponse` and return `(rsp, err)` — both non-nil — so callers can recover the partial output. On other failures, surface stderr in the returned error.
- `cmd/crossplane/render/engine_docker.go`: read the stderr return from `RunContainer` and use the new `*ContainerExitError` to detect the pipeline-fatal exit code, with the same `(rsp, err)` recovery shape on exit 3. Adds an internal `runContainer` field defaulting to `docker.RunContainer` so unit tests can inject a fake without standing up real Docker.

#### Tests added

- `cmd/crossplane/render/engine_local_test.go`: helper-process pattern (re-execs the test binary as the "binary" via `TestMain`). Cases: success, fatal-with-partial-output, fatal-no-partial-output, hard-fail with stderr.
- `cmd/crossplane/render/engine_docker_test.go`: injects a fake `runContainer`. Cases: success, fatal-with-partial-output, fatal-no-partial-output, hard-fail with `*ContainerExitError`, hard-fail with non-exit error (e.g. image pull failure).

#### Motivating consumer

`crossplane-contrib/crossplane-diff` runs the engine programmatically and needs to (a) surface FATAL stderr to users via its own error reporting and (b) iterate on `RequiredResources` even after a pipeline FATAL — the contract crossplane/crossplane#7455 was designed to enable. It currently ships its own wrapper around the local engine to add both behaviours; once this lands the wrapper goes away. Tracking PR: [crossplane-contrib/crossplane-diff#326](https://github.com/crossplane-contrib/crossplane-diff/pull/326).

Fixes #

I have:

- [ ] ~~Linked a PR or a [docs tracking issue] to [document this change].~~ (internal engine behaviour; no user-facing docs change required)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cmd/crossplane/render/engine_local_test.go
```diff
@@ -0,0 +1,188 @@
...
+// runRenderHelper which writes the canned response and exits with the
+// chosen code — never reaching m.Run(). The parent reads the result as if
+// it had just exec'd a real `crossplane internal render`. When the env
+// var is unset (the normal `go test` invocation) TestMain falls through
+// to m.Run() and the tests behave like any other Go tests.
+//
+// The cost is a few extra lines of TestMain plumbing; the benefit is a
+// fully self-contained, deterministic, cross-platform fake binary with
+// access to the same proto types the tests assert on.
+const envHelperMode = "GO_HELPER_LOCAL_RENDER_MODE"
```
jcogilvie, line 66: I was a little suspect of this, but apparently this is how go's `exec` tests itself, so the pattern is established.

jcogilvie:
@coderabbitai review

cmd/crossplane/render/engine_docker.go
```diff
@@ -34,6 +34,24 @@ import (
 	renderv1alpha1 "github.com/crossplane/cli/v2/proto/render/v1alpha1"
 )
 
+// containerRunner is the subset of internal/docker the engine depends on. It
+// follows the same one-method-interface mock pattern as pullClient in
+// runtime_docker.go, letting tests substitute a mockContainerRunner without a
+// real Docker daemon.
```
adamwg: Nit: I would keep just the first sentence of this comment :-). The rest is likely to go stale.

cmd/crossplane/render/engine_docker.go
```diff
@@ -116,8 +149,35 @@ func (e *dockerRenderEngine) Render(ctx context.Context, req *renderv1alpha1.Ren
...
+		}
+		// On a *ContainerExitError, err.Error() already embeds stderr
+		// (ContainerExitError.Error stringifies it as "container exited with
+		// status N: <stderr>"), so wrapping err is sufficient. For non-exit
+		// failures (e.g. image pull errors), append the captured stderr
+		// buffer so its content isn't lost when err.Error() doesn't include
+		// it.
+		if errors.As(err, &exitErr) {
+			return nil, errors.Wrap(err, "cannot run crossplane internal render in Docker")
+		}
```
adamwg: Could we handle this case in the block above, so we're not checking the same `errors.As` condition twice?

cmd/crossplane/render/engine_docker.go
```diff
@@ -116,8 +149,35 @@ func (e *dockerRenderEngine) Render(ctx context.Context, req *renderv1alpha1.Ren
...
+		if len(stderr) > 0 {
+			return nil, errors.Errorf("cannot run crossplane internal render in Docker: %s: %s", err.Error(), stderr)
+		}
 		return nil, errors.Wrap(err, "cannot run crossplane internal render in Docker")
```
adamwg: For readability, I think we could collapse this into a single `errors.Wrapf`, which would also make it explicit if there's no stderr output:

```suggestion
		return nil, errors.Wrapf(err, "crossplane internal render returned error with output: %s", stderr)
```

cmd/crossplane/render/engine_local.go
```diff
@@ -49,18 +48,40 @@ func (e *localRenderEngine) Setup(_ context.Context, _ []pkgv1.Function) (func()
...
+			rsp := &renderv1alpha1.RenderResponse{}
+			if uerr := proto.Unmarshal(out, rsp); uerr != nil {
+				return nil, errors.Errorf("cannot unmarshal partial render response after pipeline fatal: %s: %s", uerr.Error(), stderr.String())
+			}
+			return rsp, errors.Errorf("crossplane internal render: pipeline returned fatal: %s", stderr.String())
+		}
+		if stderr.Len() > 0 {
+			return nil, errors.Errorf("cannot run crossplane internal render: %s: %s", err.Error(), stderr.String())
+		}
 		return nil, errors.Wrap(err, "cannot run crossplane internal render")
```
adamwg: Same as above, I'd do a single `Wrapf` here.

cmd/crossplane/render/engine_local_test.go
adamwg: Certainly a little complex, but the comment explains it clearly and I don't see an obviously better solution 👍

adamwg (commented):
This lgtm overall - thanks for picking it up. I've left a few small readability/simplification suggestions.

adamwg (approved):
lgtm, thanks for the updates!

PR #105, opened 2026-06-12
Add repository option to `crossplane project init`

bobh66:

Added the `--repository` option to `crossplane project init` to override the default `example.com/my-org` values.

Made the directory name truly optional.

I have: 

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `./nix.sh flake check` to ensure this PR is ready for review.
- [X] Added `backport release-x.y` labels to auto-backport this PR.

### DIFF
--- a/cmd/crossplane/project/init.go
+++ b/cmd/crossplane/project/init.go
@@ -38,9 +38,9 @@ const projectFileName = "crossplane-project.yaml"
 
 // initCmd initializes a new project.
 type initCmd struct {
-	Name       string `arg:""                       help:"The name of the new project."`
-	Directory  string `arg:""                       help:"Directory to initialize. Defaults to project name." optional:"" short:"d" type:"path"`
-	Repository string `default:"example.com/my-org" help:"Override the repository in the project file."       optional:""`
+	Name       string `arg:""                                                    help:"The name of the new project."`
+	Directory  string `help:"Directory to initialize. Defaults to project name." short:"d"                                           type:"path"`
+	Repository string `default:"example.com/my-org"                              help:"Override the repository in the project file." optional:"" short:"r"`
 }
 
 func (c *initCmd) Help() string {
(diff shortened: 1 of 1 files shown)


bobh66:
Docs PR https://github.com/crossplane/docs/pull/1108

cmd/crossplane/project/init.go
```diff
@@ -73,8 +78,8 @@ kind: Project
 metadata:
   name: %s
 spec:
-  repository: example.com/my-org/%s
-`, c.Name, c.Name)
+  repository: %s/%s
+`, c.Name, repo, c.Name)
```
adamwg: I'd prefer to just use `repo` when it's specified, rather than appending the project name. OCI defines two things: registries and repositories; the in-between of a partially-specified repository isn't really a concept.

cmd/crossplane/project/init.go
```diff
@@ -61,6 +62,10 @@ func (c *initCmd) Run(sp terminal.SpinnerPrinter) error {
 		return err
 	}
 
+	repo := strings.TrimRight(strings.TrimSpace(c.Repository), "/")
+	if repo == "" {
+		return errors.New("repository cannot be empty; set --repository to an OCI repository prefix like 'xpkg.crossplane.io/my-org'")
```
adamwg: Might be worth validating that `repo` is a valid, fully-qualified repository by calling [`name.NewRepository`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name#NewRepository).

cmd/crossplane/project/init.go
bobh66: Don't we really have three parts - the registry, the org and the repository?  Maybe the option should be registryPath or similar to include both the registry and the org?  I'd like to be able to default the repository name to the project name rather than have to specify it again in the registry path.

adamwg: The first component of the repository being an organization name is just a convention, not specified by OCI. That said, it is a very common convention - every public registry I can think of follows it (GHCR, Docker Hub, Upbound marketplace, etc.).

Agreed it's a nice UX to not require repeating the project name. I think making the argument `--registry` (not technically correct, but less surprising to append to it) or `--repository-prefix` would make sense. That would allow for a `--repository` flag to provide the full path as well, if users want the repository name to not match the project name.

adamwg (approved):
LGTM!

adamwg:
I'm going to remove the backport label, since this isn't strictly a bug fix and we're going to release CLI v2.4 right away anyway.

PR #107, opened 2026-06-13
fix(render): support repeating --required-resources and --extra-resources

### LINKED ISSUE #90: `crossplane render` only honors the LAST `--extra-resources` flag
## Summary

`crossplane render` (and `crossplane composition render`) **silently ignores all but the
last `--extra-resources` / `--required-resources` flag**. When the flag is passed more
than once, only the file/dir given in the final occurrence is loaded; resources in the
earlier occurrences are never made available to the function pipeline — with no error or
warning. Functions that request those resources receive an empty set.

- CLI: `crossplane` v2.3.1
- Render engine: `--crossplane-version=v2.3.1` (Docker-based render; default `stable`)
- function-go-templating v0.12.1, function-auto-ready v0.6.5

## Files

- `xr.yaml` — a trivial composite resource
- `xrd.yaml` — the XRD defining the `XR` composite type. **Optional** — `crossplane
  render` does not require an XRD (it renders the XR directly against the Composition);
  it's included only to make the example self-contained and can be passed with `--xrd`.
- `min-composition.yaml` — one pipeline step (`function-go-templating`) that requests a
  ConfigMap labelled `group=a` and writes the number it received into the composed
  ConfigMap's `data.count`
- `min-wanted.yaml` — the ConfigMap the function wants (`group=a`)
- `min-filler.yaml` — an unrelated ConfigMap (`group=b`)
- `functions-public.yaml` — same functions using public `xpkg.crossplane.io` images

[functions-public.yaml](https://github.com/user-attachments/files/28723259/functions-public.yaml)
[min-composition.yaml](https://github.com/user-attachments/files/28723256/min-composition.yaml)
[min-filler.yaml](https://github.com/user-attachments/files/28723257/min-filler.yaml)
[min-wanted.yaml](https://github.com/user-attachments/files/28723261/min-wanted.yaml)
[xr.yaml](https://github.com/user-attachments/files/28723258/xr.yaml)
[xrd.yaml](https://github.com/user-attachments/files/28723260/xrd.yaml)

## Run

```bash
# Docker must be running (the engine renders in a container)

# A) single flag — works, count == 1
crossplane composition render --crossplane-version=v2.3.1 \
  xr.yaml min-composition.yaml functions-public.yaml \
  --extra-resources min-wanted.yaml | grep count

# B) two flags, wanted THEN filler — BUG: count == 0 (wanted file is dropped)
crossplane composition render --crossplane-version=v2.3.1 \
  xr.yaml min-composition.yaml functions-public.yaml \
  --extra-resources min-wanted.yaml \
  --extra-resources min-filler.yaml | grep count

# C) two flags, filler THEN wanted — works, count == 1 (only the LAST flag is honored)
crossplane composition render --crossplane-version=v2.3.1 \
  xr.yaml min-composition.yaml functions-public.yaml \
  --extra-resources min-filler.yaml \
  --extra-resources min-wanted.yaml | grep count
```

## Observed

| Invocation | `data.count` |
|---|---|
| A) `--extra-resources wanted` | `"1"` ✅ |
| B) `--extra-resources wanted --extra-resources filler` | `"0"` ❌ |
| C) `--extra-resources filler --extra-resources wanted` | `"1"` ✅ |

Putting both resources in a **single** file (or a single directory) also yields `"1"`,
confirming the data is fine — only the repeated *flag* is mishandled.

The same is observable with `--required-resources` (the non-deprecated flag) and with a
mix of resource kinds (e.g. a step requesting a custom resource plus a
`function-environment-configs` step requesting an `EnvironmentConfig`): whichever
requirement's resource is not in the final `--extra-resources` file is dropped, and the
corresponding function silently composes nothing.

## Expected

Resources from **all** `--extra-resources` / `--required-resources` occurrences should be
merged and made available to the pipeline.

## Workaround

Pass a **single** `--extra-resources` argument — either one YAML file containing all the
resources, or a **directory** of YAML files.

chaitanyapantheor:
## Summary

Fixes #90

`--required-resources` and `--extra-resources` (deprecated) on `crossplane render xr` and `crossplane render op` silently ignored all but the **last** occurrence when the flag was repeated. Resources from earlier flags were dropped with no error or warning, causing functions that needed those resources to receive an empty set.

**Root cause:** The fields were declared as `string` instead of `[]string`. Kong's CLI parser silently overwrites a `string` field on repeated flags, keeping only the last value.

**Fix:** Change both flags to `[]string` and loop over all provided paths in `Run()`, merging resources from every file/directory into a single slice.

## Test Plan

- [x] Added `MultipleRequiredResourcesFirstMissing` test to `render xr` — verifies an error on the first of two files propagates (previously silently dropped)
- [x] Added `MultipleExtraResourcesFirstMissing` test to `render xr` — same for deprecated `--extra-resources`
- [x] Added `MultipleRequiredResourcesFirstMissing` test to `render op`
- [x] Updated existing `LoadRequiredResourcesError` tests to use `[]string`
- [x] Full test suite passes (`go test ./...`)

## Example

```bash
# Before: only resources-b.yaml was loaded; resources-a.yaml was silently dropped
crossplane render xr xr.yaml composition.yaml functions.yaml \
  --required-resources resources-a.yaml \
  --required-resources resources-b.yaml

# After: both files are loaded and merged
crossplane render xr xr.yaml composition.yaml functions.yaml \
  --required-resources resources-a.yaml \
  --required-resources resources-b.yaml
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adamwg (approved):
I think this is going to have some lint errors due to struct tag alignment, but LGTM once that's fixed. Thanks for the PR, @chaitanyapantheor!

chaitanyapantheor:
> I think this is going to have some lint errors due to struct tag alignment, but LGTM once that's fixed. Thanks for the PR, @chaitanyapantheor!

@adamwg have addressed them, hoping this should resolve the all lint issues 🖕

PR #114, opened 2026-06-15
feat: add --as/--as-group/--as-uid impersonation flags

### LINKED ISSUE #110: Implement privilege elevation flag

Currently, there's no way to run `crossplane` CLI commands as a different user or service account. `kubectl` has `--as` and `--as-group` for this, but the crossplane CLI doesn't have anything equivalent.

This comes up when:

- You want to check if a specific user can actually see or access a resource (e.g. with `resource trace`) without switching your whole kubeconfig context
- You're debugging access issues for another user and want to see what they see
- You're running in CI as a privileged service account but need to validate things from the perspective of a less-privileged one
- You're working in a multi-tenant cluster and want to confirm resources are scoped correctly for a given tenant

Add `--as`, `--as-group`, and optionally `--as-uid` flags to the CLI commands that talk to the cluster.

A few thoughts on how this could work:

- **Which commands**: `resource trace`, `cluster top`, `version`, `xpkg install`, `xpkg update`, and the completion helpers all create a Kubernetes client, so they'd all benefit from these flags.
- **Shared definition**: Rather than duplicating the flags in every command, it'd be nice to define them once in a shared struct and mix them in where needed.
- **Match kubectl**: The flag names and behavior should match what `kubectl` does so there's nothing new to learn.

Example usage:

```bash
# Trace a resource as a specific user
crossplane resource trace compositeresource.example.org my-resource --as=jane@example.com

# Check cluster top as a service account
crossplane cluster top --as=system:serviceaccount:team-a:reader

# Install a package impersonating a group
crossplane xpkg install provider example/provider-foo --as-group=team-a-admins
```

mikeshootzz:

Adds the kubectl-compatible impersonation flags as, as-group, and as-uid to every CLI command that talks to a cluster. This lets you run a command as a different user, group, or service account without switching your kubeconfig context.

The flags are defined once in a shared ImpersonationFlags struct and embedded into each command. After a command builds its client config, it applies the flags, which sets the impersonation fields on the request. This works for every command regardless of how it loads its config.

Commands covered: resource trace, cluster top, version, xpkg install, xpkg update, and the shell-completion predictors.

The behavior matches kubectl: the flag names and help text are the same, as-group is repeatable, there are no short forms, and validation is left to the API server.

Fixes #110 

I have: 

- [ ] ~~Linked a PR or a [docs tracking issue] to [document this change].~~
- [ ] ~~Added `backport release-x.y` labels to auto-backport this PR.~~


cmd/crossplane/trace/trace_test.go
```diff
@@ -103,3 +104,23 @@ func TestCmd_getResourceAndName(t *testing.T) {
...
+		t.Fatalf("Parse(): unexpected error: %v", err)
+	}
+
+	if c.Impersonation.As != "jane" {
+		t.Errorf("As: want %q, got %q", "jane", c.Impersonation.As)
+	}
+	if len(c.Impersonation.AsGroup) != 1 || c.Impersonation.AsGroup[0] != "team-a" {
+		t.Errorf("AsGroup: want [team-a], got %v", c.Impersonation.AsGroup)
+	}
+}
```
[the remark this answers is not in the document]
mikeshootzz: This should be fixed now

internal/kube/impersonation.go
adamwg, line 1: I'd prefer to put this functionality in `internal` rather than `cmd/crossplane/common`. It's CLI-specific, so I don't see any reason for it to be importable by external codebases, and we should get rid of this `common` directory eventually (see below).

This directory is a leftover from the CLI being in the core crossplane/crossplane repository (which intentionally doesn't have a `pkg/` for exported packages) and wanting to expose some code for CLI-adjacent utilities like crossplane-diff to use. We'd like to start moving the code that lives here into either `internal/` or `pkg/` as appropriate to make it clear what can/should be imported externally (it's a bit of a smell to import code from `cmd/`).

cmd/crossplane/top/top_test.go
```diff
@@ -7,15 +7,56 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	v1 "github.com/crossplane/crossplane/apis/v2/pkg/v1"
+
+	"github.com/crossplane/cli/v2/cmd/crossplane/common/kube"
 )
 
+func TestImpersonationFlagsParse(t *testing.T) {
```
adamwg: I'm not sure this test is very valuable - it's mostly testing the internals of kong rather than our code.

If you do want to keep it, I'd suggest putting it alongside the `ImpersonationFlags` struct in the `kube` package rather than duplicating it in each command that uses the flags.

adamwg (changes requested):
I've left a couple of notes on code organization and tests, but the actual implementation here looks great. Thanks for the contribution, @mikeshootzz!

internal/kube/impersonation.go
mikeshootzz: Thanks for letting me know! I'll move it to internal later today.

cmd/crossplane/top/top_test.go
mikeshootzz: Fair point. I'll fix that. Thanks!

adamwg (approved):
LGTM, thanks for the contribution!

adamwg:
Fixed the vale issue in the help text so we can get this merged for a v2.4 release :pray:

PR #119, opened 2026-06-15
Drop the scale subresource before building OpenAPI for schema generation

### LINKED ISSUE #117: crossplane xrd generate models the autoscaling Scale type instead of the resource when a scale subresource is declared
### What happened?

When an XRD (or CRD) declares a `scale` subresource, `crossplane xrd generate` produces language schemas for the `autoscaling/v1` `Scale` type instead of for the resource itself. The resource's own model is replaced entirely.

Crossplane v2.3 added the ability to configure the scale subresource on an XRD ([crossplane/crossplane#7004](https://github.com/crossplane/crossplane/pull/7004)). Adding it to an XRD like this:

```yaml
versions:
- name: v1alpha1
  subresources:
    scale:
      specReplicasPath: .spec.replicas
      statusReplicasPath: .status.replicas.total
```

causes the generated Python module for the resource (e.g. `models/ai/modelplane/modeldeployment/v1alpha1.py`) to contain only `Scale`, `ScaleSpec`, and `ScaleStatus` classes. The resource's own kind, spec, and status classes are gone, and the module's relative imports change shape, so anything importing the generated model fails:

```
ModuleNotFoundError: No module named 'models.ai.apimachinery'
```

### Root cause

`internal/crd/convert.go`'s `ToOpenAPI` builds the intermediate OpenAPI document with `builder.BuildOpenAPIV3` (from `k8s.io/apiextensions-apiserver`). That builder emits the full Kubernetes REST surface for the resource: `paths` for list/get/create/**scale**, and a `components.schemas` map.

When the resource has a scale subresource, the builder adds `io.k8s.api.autoscaling.v1.Scale`, `ScaleSpec`, and `ScaleStatus` to `components.schemas` alongside the resource's own schema. The downstream language generator (datamodel-code-generator for Python) then models those `Scale` schemas into the resource's module, displacing the resource.

You can see it directly: feeding a CRD with a scale subresource through `crd.FilesToOpenAPI` yields a `components.schemas` map containing both `ai.<group>.<version>.<Kind>` and `io.k8s.api.autoscaling.v1.Scale` (plus `ScaleSpec`/`ScaleStatus`). Clearing the scale subresource before the conversion removes the autoscaling schemas and leaves the resource's schema intact.

### How can we reproduce it?

1. Add a scale subresource to any XRD version (the YAML above).
2. Run `crossplane xrd generate` (or `crossplane dependency update-cache` / `crossplane project build`, which generate schemas) with `schemas.languages: [python]`.
3. Inspect the generated module for that resource: it contains `Scale`/`ScaleSpec`/`ScaleStatus` instead of the resource's own classes.

* Crossplane CLI version: `main` (commit `3df82fc`)
* Platform: linux/amd64
* Crossplane version (if applicable): v2.3.2

negz:
Fixes #117

When an XRD or CRD declares a `scale` subresource, `crossplane xrd generate` produces language schemas for the `autoscaling/v1` `Scale` type in place of the resource's own model. For Python the generated module's imports also change shape, so importing the model fails outright with `ModuleNotFoundError: No module named 'models.ai.apimachinery'`.

`ToOpenAPI` builds the intermediate OpenAPI document with `BuildOpenAPIV3`, which models the resource's whole REST surface, not just its structural schema. Declaring a scale subresource adds a `/scale` endpoint whose request and response body is an `autoscaling/v1` `Scale`, so the builder adds `Scale`, `ScaleSpec`, and `ScaleStatus` to the document's components. The meta types it always emits (`ObjectMeta` and friends) get redirected to a shared module by a later transform, but the autoscaling types don't, so the language generator models them into the resource's own module.

The scale subresource is a runtime API behaviour with no bearing on the generated model, so this clears it before the conversion. `BuildOpenAPIV3` then never emits the `/scale` endpoint or its `Scale` schemas, and the resource's model generates as it did before.

I have:

- [x] Run `nix flake check` to ensure this PR is ready for review.
- [ ] ~Linked a PR or a docs tracking issue to document this change.~

### DIFF
--- a/internal/crd/convert.go
+++ b/internal/crd/convert.go
@@ -128,6 +128,17 @@ func modifyCRDManifestFields(crd *extv1.CustomResourceDefinition) {
 			updateSchemaPropertiesXEmbeddedResource(version.Schema.OpenAPIV3Schema)
 			crd.Spec.Versions[i].Schema.OpenAPIV3Schema.Properties = version.Schema.OpenAPIV3Schema.Properties
 		}
+
+		// The scale subresource is a runtime API behaviour, not part of the
+		// resource's structural schema. But BuildOpenAPIV3 models the whole REST
+		// surface, so declaring it adds a /scale endpoint whose request and
+		// response body is an autoscaling/v1 Scale. That pulls the Scale,
+		// ScaleSpec, and ScaleStatus schemas into the resource's components,
+		// which the language generators then model in place of the resource
+		// itself. Drop it before building the OpenAPI spec.
+		if version.Subresources != nil {
+			crd.Spec.Versions[i].Subresources.Scale = nil
+		}
 	}
 }
 
--- a/internal/crd/convert_test.go
+++ b/internal/crd/convert_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package crd
 
 import (
+	"sort"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


adamwg (approved):
LGTM. It seems like the scale subresource causes a problem only for Python (I believe due to how we structure the generated Python packages), but dropping scale before generating is reasonable.

I wonder if this would let us also simplify some of the generator code, since I know it has special handling for the v1/autoscaling Scale resource.

PR #127, opened 2026-06-18
Decompress function runtime tarballs once when loading

negz:

#24 added support for gzipped function runtime image tarballs by streaming each one through `gzip.NewReader` directly into go-containerregistry's `tarball.Image`.

go-containerregistry calls the `tarball.Opener` it's given once per layer, plus once each for the manifest and config. Because the gzip opener re-opened and re-decompressed the whole file from the start on every call, loading a single image decompressed it once per layer.

My project uses Nix's `dockerTools`, which emits one layer per store path, so a typical function image has ~50 layers and was fully gunzipped ~54 times. Computing the image digest then re-reads every layer again. With functions built concurrently, a project with a dozen multi-arch functions spent over ten minutes pegging every core in this loop.

This PR decompresses each gzipped tarball once into a temporary file and serves every opener call from that plain tar. The temporary files back the returned images lazily, so they must outlive `Build`; the builder now creates them under a per-build temporary directory and exposes a `Close` method that removes the directory once the caller has finished consuming the images. `NewBuilder` returns the concrete `*realBuilder` so callers can `defer Close`, and the build, run, and render entry points do so after they have written, sideloaded, or loaded the images.

On a project with twelve functions built for amd64 and arm64, loading all twenty-four images drops from over ten minutes to roughly eighty seconds.

This follows on from #21 and #24, which introduced pre-built and gzipped function runtime tarball support respectively.

Fixes #

I have:

### DIFF
--- a/cmd/crossplane/project/build.go
+++ b/cmd/crossplane/project/build.go
@@ -19,6 +19,7 @@ package project
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -123,11 +124,21 @@ func (c *buildCmd) Run(logger logging.Logger, sp terminal.SpinnerPrinter) error 
 		dependency.WithResolver(resolver),
 	)
 
+	// The builder may decompress function runtime tarballs into this directory;
+	// the built images read from it lazily, so we remove it only after the
+	// package has been written below.
+	tempDir, err := os.MkdirTemp("", "crossplane-build-")
+	if err != nil {
+		return errors.Wrap(err, "failed to create temporary build directory")
+	}
+	defer os.RemoveAll(tempDir) //nolint:errcheck // Best-effort cleanup of temporary files.
+
 	b := project.NewBuilder(
 		project.BuildWithMaxConcurrency(concurrency),
 		project.BuildWithFunctionIdentifier(functions.DefaultIdentifier),
 		project.BuildWithSchemaManager(schemaMgr),
 		project.BuildWithDependencyManager(depMgr),
+		project.BuildWithTempDir(tempDir),
 	)
 
 	var imgMap project.ImageTagMap
--- a/internal/project/build.go
+++ b/internal/project/build.go
@@ -103,14 +103,12 @@ func BuildWithDependencyManager(m *dependency.Manager) BuilderOption {
 	}
 }
 
-// BuildWithTempDir sets the directory into which the builder decompresses
-// gzipped function runtime tarballs. The returned images may read from these
-// files lazily, so the caller owns the directory and must remove it only after
-// it has finished consuming the images.
+// BuildWithTempDir sets a directory the builder can use to hold temporary
+// files. The images returned from Build() may depend on this directory, so
+// callers should not remove it until they have finished consuming the images.
 //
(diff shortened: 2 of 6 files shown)


cmd/crossplane/project/build.go
```diff
@@ -123,11 +124,21 @@ func (c *buildCmd) Run(logger logging.Logger, sp terminal.SpinnerPrinter) error
...
 	)
 
+	// The builder may decompress function runtime tarballs into this directory;
+	// the built images read from it lazily, so we remove it only after the
+	// package has been written below.
+	tempDir, err := os.MkdirTemp("", "crossplane-build-")
+	if err != nil {
+		return errors.Wrap(err, "failed to create temporary build directory")
+	}
+	defer os.RemoveAll(tempDir) //nolint:errcheck // Best-effort cleanup of temporary files.
```
negz, line 134: I considered setting this up once in `main.go` and plumbing it down, but the plumbing felt quite a bit more awkward. I decided the repetition was the less bad approach.

internal/project/build.go
```diff
@@ -67,38 +67,29 @@ const (
 // ImageTagMap is a map of container image tags to images.
 type ImageTagMap map[name.Tag]v1.Image
 
-// Builder is able to build a project into a set of packages.
```
negz, line 70: I had originally added a `Close()` method to `Builder`, but it ended up needing to track all the tarballs it'd opened (for every `Build` call) so I switched path. Along the way I realized nothing actually cares about this interface. In the end it's a bit of a drive by refactor. Happy to revert if desired but I think it's an improvement.

cmd/crossplane/project/build.go
adamwg: Agreed, feels cleaner to have it set up in each command that needs it.

internal/project/build.go
adamwg: I think this was an interface originally to allow/simplify mocking it in tests. We've ended up making the real builder configurable enough that we can just inject its parts rather than injecting a whole builder, so the interface ended up with only a single implementation. No objection to the refactor 👍

internal/project/build.go
```diff
@@ -107,11 +98,25 @@ func BuildWithSchemaManager(m *manager.Manager) BuilderOption {
...
 }
 
+// BuildWithTempDir sets the directory into which the builder decompresses
+// gzipped function runtime tarballs. The returned images may read from these
+// files lazily, so the caller owns the directory and must remove it only after
+// it has finished consuming the images.
+//
+// If unset, the decompressed tarballs are written to the OS default temporary
+// directory and are not removed; callers that build gzipped tarball functions
+// should set this to a directory they delete.
```
adamwg: Nit: this comment is pretty specific to the current usage of the temp dir, which I can imagine being used by other parts of the build in the future. I'd probably make it more general to reduce the chances of it becoming stale.

```suggestion
// BuildWithTempDir sets a directory the builder can use to hold temporary
// files. The images returned from Build() may depend on this directory, so
// callers should not remove it until they have finished consuming the images.
//
// If unset, the OS default temporary directory is to hold temporary files,
// which are not cleaned up by the builder.
```

adamwg (approved):
One comment inline, but this lgtm overall.

PR #130, opened 2026-06-18
schemas: Prime oapi-codegen's global state to make capitalization consistent

### LINKED ISSUE #113: Generated Go schemas use inconsistent capitalization
### What happened?

Normally, when I generate Go schemas for the Kubernetes types (e.g., with `crossplane dependency add k8s:v1.35.0`), I get the following const generated:

```go
// Defines values for ServiceAPIVersion.
const (
        ServiceAPIVersionV1 ServiceAPIVersion = "v1"
)
```

But sometimes the capitalization is different, which breaks compilation of any functions that use the const:

```go
// Defines values for ServiceApiVersion.
const (
        ServiceApiVersionV1 ServiceApiVersion = "v1"
)
```

### How can we reproduce it?

I'm not able to reproduce this consistently - most of the time the expected `API` is generated. However, I have seen the incorrect `Api` pop up more than once.

* Crossplane CLI version: v2.3.2
* Crossplane version (if applicable): n/a

adamwg:

The oapi-codegen package we use to generate Go code keeps a bunch of internal global state that affects how code is generated. In particular, the name normalizer it uses to form types names is a global variable initialized when we call `codegen.Generate`.

We call some of the codegen package's utility functions to form custom names for types before generating code. These utility functions also use the global state set up by `codegen.Generate`, which isn't initialized when we call them, resulting in the utility functions producing different names than we expect. This makes our code generation non-deterministic: we get different name capitalization depending what order schemas get generated in.

Make a no-op call to `codegen.Generate` with our desired options before calling the mutators to ensure that the global state is initialized for all our calls. Note that once https://github.com/oapi-codegen/oapi-codegen/pull/2393 is merged in the upstream code generator, we can use the new `SetGlobalStateOptions` function rather than the dummy `Generate` call.

Fixes #113

I have: 

### DIFF
--- a/internal/schemas/generator/go.go
+++ b/internal/schemas/generator/go.go
@@ -377,6 +377,39 @@ func goCollectOpenAPIs(fromFS afero.Fs) ([]goOpenAPI, error) { //nolint:gocognit
 var generateGoMutex sync.Mutex //nolint:gochecknoglobals // Must be global.
 
 func generateGo(s *spec3.OpenAPI, version string, mutators ...func(*spec3.OpenAPI)) (string, error) {
+	// codegen.Generate sets some global state that's used by the utility
+	// functions we call from our mutators. That has two implications for us:
+	//
+	// 1. We must hold the `generateGoMutex` while calling the mutators, not
+	//    just when we call `codegen.Generate`.
+	// 2. We must call `codegen.Generate` with the options we're going to use
+	//    *before* we call the mutators, so that the global state is correct
+	//    inside the codegen package. We call it with a minimal input and ignore
+	//    the output - this call is just to set up the global state. Once
+	//    https://github.com/oapi-codegen/oapi-codegen/pull/2393 is merged we
+	//    can use `codegen.SetGlobalStateOptions` instead of calling `Generate`.
+	generateGoMutex.Lock()
+	defer generateGoMutex.Unlock()
+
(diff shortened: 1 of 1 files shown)


internal/schemas/generator/go.go
```diff
@@ -542,9 +560,6 @@ func goRenamePropertyTypes(baseName string, props map[string]spec.Schema) {
 }
 
 func goFixName(name string) string {
-	generateGoMutex.Lock()
```
jbw976, line 545: can you help confirm why this mutex was needed before, but isn't now? i imagine that it's because we hold it earlier in the `generateGo` function now, so we don't need it here, but i also would have expected to see other functions affected in the same way 🤔

jbw976 (approved):
looks like a reasonable refactor, just a question about the mutex removal from goFixName

internal/schemas/generator/go.go
adamwg: Good question! This is the only mutator that requires we hold the lock. Previously we acquired (and released) it here, then acquired and released it again around the call to `codegen.Generate()` in `generateGo`. Now we have two calls to `codegen.Generate()` in `generateGo` with the mutator calls between them, so I opted to take the mutex once for all of `generateGo` rather than acquiring and releasing it three times.

PR #160, opened 2026-06-26
feat: add GetX/SetX accessors to generated Go models

erikmiller-gusto:
Add accessor/setter methods to generated Go models

# Problem

The Go model generator produces structs that expose only fields:
```
type XAccountScaffoldSpec struct {
    Parameters *XAccountScaffoldSpecParameters `json:"parameters,omitempty"`
    // ...
}
```
Because the generated types have no methods, consumers can't abstract over resources with interfaces or write generic code that operates on "any resource with field X". Every field is a pointer (the goRemoveRequired mutator strips required, so oapi-codegen emits everything as optional *T).

# What this does

Generates GetX/SetX accessor methods for every field of every generated struct, so consumers can define their own interfaces and satisfy them structurally:

```
func (o *XAccountScaffold) GetSpec() *XAccountScaffoldSpec  { return o.Spec }
func (o *XAccountScaffold) SetSpec(v *XAccountScaffoldSpec) { o.Spec = v }

// consumer code:
type SpecAccessor interface {
    GetSpec() *v1alpha1.XAccountScaffoldSpec
    SetSpec(*v1alpha1.XAccountScaffoldSpec)
}
var _ SpecAccessor = &v1alpha1.XAccountScaffold{}
```
# Testing

- Unit test over *string / *[]T / *map[...] / *Struct fields, including alias-skipping.
- Integration test asserting real CRD models and shared k8s ObjectMeta get accessors.
- Compile gate: materializes the generated module to disk, adds a consumer that uses an accessor through an interface, and go builds it — proving the output compiles and the goal (interfaces over resources) actually works. Skips gracefully if no Go toolchain is present.

go test ./... → all passing. go build, gofmt, go vet clean.

I have: 

On the two struck items: the `crossplane config` help text this PR extends is
rendered into the generated command reference, so `crossplane/docs` picks the new
key up without a separate change. And this is a new opt-in feature rather than a
fix, so it isn't a backport candidate.

### DIFF
--- a/cmd/crossplane/config/help/config.md
+++ b/cmd/crossplane/config/help/config.md
@@ -18,3 +18,10 @@ Enable alpha commands:
 ```shell
 crossplane config set features.enableAlpha true
 ```
+
+Generate GetX/SetX accessor methods on generated Go models (off by default), so
+generated resources can be used through interfaces and generics:
+
+```shell
+crossplane config set features.generateGoModelAccessors true
+```
--- a/cmd/crossplane/config/help/config.md
+++ b/cmd/crossplane/config/help/config.md
@@ -20,7 +20,7 @@ crossplane config set features.enableAlpha true
 ```
 
 Generate GetX/SetX accessor methods on generated Go models (off by default), so
-generated resources can be used through interfaces and generics:
+you can reach generated resources through interfaces and generics:
 
 ```shell
 crossplane config set features.generateGoModelAccessors true
--- a/cmd/crossplane/main.go
+++ b/cmd/crossplane/main.go
@@ -126,6 +126,8 @@ func main() {
 		// at runtime.
 		kong.BindTo(logger, (*logging.Logger)(nil)),
 		kong.BindTo(configcmd.ConfigPath(cfgPath), (*configcmd.ConfigPath)(nil)),
+		// Bind the loaded config so commands can read feature flags at runtime.
+		kong.Bind(cfg),
 		kong.Help(helpPrinter),
 		kong.UsageOnError())
 
--- a/internal/schemas/generator/accessors.go
+++ b/internal/schemas/generator/accessors.go
@@ -31,6 +31,16 @@ import (
 // alias, which are all multi-letter.
 const accessorReceiver = "o"
 
+// applyAccessors runs addAccessors when enabled. It is called by the generation
+// loops after all other Go post-processing, so the accessors reference the
+// final, fixed-up type names.
+func applyAccessors(code string, enabled bool) (string, error) {
+	if !enabled {
+		return code, nil
+	}
+	return addAccessors(code)
+}
+
 // addAccessors generates GetX/SetX accessor methods for every field of every
 // struct type declared in the given Go source. Getters return the field's
 // (pointer) type as-is and setters take the same type, so the generated methods
@@ -139,6 +149,13 @@ func writeStructAccessors(b *strings.Builder, fset *token.FileSet, typeName stri
 		fieldType := typ.String()
 
 		for _, name := range field.Names {
+			// Skip unexported fields: an accessor for them would be useless to
+			// external consumers and could produce oddly-cased method names.
+			// Generated models don't currently have any, but guard defensively.
+			if !name.IsExported() {
+				continue
+			}
+
 			fieldName := name.Name
 
 			// Getter.
--- a/internal/schemas/generator/accessors.go
+++ b/internal/schemas/generator/accessors.go
@@ -1,0 +1,161 @@
+/*
+Copyright 2026 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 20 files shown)


cmd/crossplane/main.go
```diff
@@ -126,6 +126,8 @@ func main() {
 		// at runtime.
 		kong.BindTo(logger, (*logging.Logger)(nil)),
 		kong.BindTo(configcmd.ConfigPath(cfgPath), (*configcmd.ConfigPath)(nil)),
+		// Bind the loaded config so commands can read feature flags at runtime.
+		kong.Bind(cfg),
```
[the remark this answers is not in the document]
adamwg: Kong will panic if a `Run` method takes an argument of a type that hasn't been bound, so adding defensive nil checks is not especially helpful. We don't have a great test suite yet, but once we have even basic e2e testing we'll catch these kinds of errors in CI anyway, and a panic is appropriate for programming errors.

cmd/crossplane/config/help/config.md
```diff
@@ -18,3 +18,10 @@ Enable alpha commands:
 ```shell
 crossplane config set features.enableAlpha true
 ```
+
+Generate GetX/SetX accessor methods on generated Go models (off by default), so
+generated resources can be used through interfaces and generics:
```
adamwg: Seems reasonable to disable this by default initially in case some bug in the generation results in broken Go bindings for some cases. I do think we should plan to enable it by default at some point, so that there's only one code path to test. We haven't really defined what the feature lifecycle looks like for the CLI, but maybe for now we can just create an issue to remind us to come back to this a couple releases down the line.

internal/schemas/generator/accessors.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return ""
+}
+
+// writeStructAccessors appends a getter and setter for each named field of the
+// given struct to b. Any accessor whose name already exists in skip is omitted
+// to avoid colliding with methods oapi-codegen already generated.
+func writeStructAccessors(b *strings.Builder, fset *token.FileSet, typeName string, st *ast.StructType, skip map[string]bool) {
+	for _, field := range st.Fields.List {
+		// Skip embedded/anonymous fields; generated models don't use them.
```
adamwg, line 138: Should we also skip unexported fields? Not sure there actually are any in our generated models, but seems like a good safeguard.

internal/schemas/generator/accessors_test.go
```diff
@@ -0,0 +1,393 @@
...
+			typ = renderExpr(t, fn.Type.Results.List[0].Type)
+		case fn.Type.Params != nil && len(fn.Type.Params.List) == 1:
+			typ = renderExpr(t, fn.Type.Params.List[0].Type)
+		}
+		out[recv+"."+fn.Name.Name] = typ
+	}
+	return out
+}
+
+func renderRecv(e ast.Expr) string {
```
adamwg: This is effectively the same as `receiverTypeName` in `accessors.go`. Any reason to redefine it just for tests? Seems equally likely to have a bug in either definition :-).

internal/schemas/generator/go.go
```diff
@@ -376,7 +380,7 @@ func goCollectOpenAPIs(fromFS afero.Fs) ([]goOpenAPI, error) { //nolint:gocognit
 // `generateGo`, since `codegen.Generate` is not concurrency safe.
 var generateGoMutex sync.Mutex //nolint:gochecknoglobals // Must be global.
 
-func generateGo(s *spec3.OpenAPI, version string, mutators ...func(*spec3.OpenAPI)) (string, error) {
+func generateGo(s *spec3.OpenAPI, version string, accessors bool, mutators ...func(*spec3.OpenAPI)) (string, error) {
```
adamwg: Would it make sense to handle the accessor generation in this PR and the `runtime.Object` method generation in #162 the same way? I.e., either pass a set of "code mutators" to `generateGo` and make it responsible for adding both sets of extra methods, or have the calling code generate the additional methods after the fact as in #162.

Totally fair if there's a reason that PR and this PR work differently, I was just surprised by it :-).

adamwg (commented):
Thanks for the contribution @erikmiller-gusto!

The core of this looks good to me. Leaving a few thoughts inline. My biggest concern is unifying the approach with #162 if possible, preferably also setting us up nicely to make the additional code generation non-optional in the future with minimal changes.

adamwg (commented):
The code generation here looks good to me, and works in my quick testing. Would be great for @haarchri to take a look as well, since he's worked in the Go generator code a lot.

One issue I did notice in testing: threading through the config option is missing in a few places where we generate schemas. Specifically, the `crossplane dependency` commands (`add` and `update-cache`) and the `render` commands (both `composition` and `operation`). The latter are a little tricky since they currently use the default generator from `dependency.NewManager`.

haarchri:
Overall LGTM one thing to clarify, when we using schemas from the official Providers we missing these new features from this PR and #162

adamwg:
> Overall LGTM one thing to clarify, when we using schemas from the official Providers we missing these new features from this PR and #162

True, for `up` users. For `crossplane` CLI users schema generation is always client-side.

haarchri:
means we starting to have breaking Changes between up and crossplane - something to discuss internally

haarchri (approved):

adamwg (approved):
LGTM, thanks for all your work on this @erikmiller-gusto!

PR #162, opened 2026-06-26
generator: make generated Go types implement runtime.Object and provide AddToScheme

### LINKED ISSUE #143: Make generated Go types implement `runtime.Object` and provide package-level `AddToScheme`s

When writing Go functions using the kubebuilder Go structs from providers, it's possible to add add API types to a `runtime.Scheme`, which avoids the need to set the `APIVersion` and `Kind` fields. This depends on the types implementing `runtime.Object`.

Our generated Go schemas don't implement `runtime.Object`, and we don't provide `AddToScheme` helpers in the generated packages. This means users have to specify `APIVersion` and `Kind` for each composed resource, and it's also not possible to use the generated types with k8s ecosystem libraries that work with `runtime.Object`s.

Slack ref: https://crossplane.slack.com/archives/CEG3T90A1/p1782299819029779

Generate the necessary methods to make our generated Go types implement `runtime.Object`, and generate `AddToScheme` helpers in each generated package.

erikmiller-gusto:

Fixes #143.

## What

Generates `runtime.Object` support for the Go schema models, behind a new opt-in
feature flag. When enabled, generated Go models can be registered in a
`runtime.Scheme` and used with k8s ecosystem libraries, so users no longer have
to set `apiVersion`/`kind` by hand on composed resources.

For **every** generated struct:
- controller-gen-style `DeepCopyInto` / `DeepCopy`.

For **root types** (structs with `APIVersion` + `Kind` + `Metadata`, i.e. the
resource and its `List`):
- `DeepCopyObject() runtime.Object`
- `GetObjectKind`/`GroupVersionKind`/`SetGroupVersionKind` (the type implements
  `schema.ObjectKind`, reading/writing the typed `APIVersion`/`Kind` fields)
- an `init()` registering the type with the package `SchemeBuilder`.

Per package containing root types, a `groupversion_info.go` defining
`GroupVersion`, `SchemeBuilder` and `AddToScheme` (apimachinery-only; no
controller-runtime dependency).

## Feature gate

Off by default; enable with:

```shell
crossplane config set features.generateGoRuntimeObjects true
```

The flag is threaded from config through every command that generates schemas:
`project build`, `project run`, `function generate`, `dependency add`,
`dependency update-cache`, `composition generate`, `composition render` /
`render`, and `operation render`. The render commands also hand the same
generators to the dependency manager, so dependency schemas match the project's
own. `dependency clean-cache` is the one exception — it only removes generated
schemas, so it keeps the flag-off default in `dependency.NewManager`.

When the flag is off, no `runtime.Object`, `DeepCopy` or `groupversion_info.go`
code is emitted.

## Dependencies

The generated models module requires `k8s.io/apimachinery`, pinned to **v0.33.0**
to match the function Go template — so a generated function that consumes the
models via `replace` still resolves everything from the template's existing
`go.sum` (verified locally by building a function against the models using the
shipped template `go.mod`/`go.sum`).

There is one `go.mod`/`go.sum` regardless of the flag. With the flag off nothing
imports apimachinery, but the requirement stays: an unused requirement is
harmless to `go build`, `go mod tidy` removes it for anyone who cares, and one
set of module files is one thing fewer to maintain and test.

## Testing

- Unit tests for the DeepCopy / runtime.Object generation (root vs non-root
  detection, scalar/struct/alias field handling), for the API group written into
  `groupversion_info.go` for built-in Kubernetes packages, and for the feature
  flag plumbing.
- A default-off test confirming no `runtime.Object` artifacts are emitted.
- **Compile gates** that materialize the generated module (the CRD path, the
  OpenAPI path, the flag-off path, and both features on) and `go build` /
  `go test` it — these
  catch `DeepCopyInto` codegen bugs that parse cleanly but don't type-check. One
  of them is a **behavioral** test asserting deep-copy independence for scalar,
  slice-of-struct and map fields, `AddToScheme` GVK round-tripping, and that
  `SetGroupVersionKind` writes the typed fields.

  The compile gates shell out to the Go toolchain to resolve the generated
  module's dependencies, so they need network access and a writable module
  cache. The Nix sandbox our unit tests run in has neither, and the generated
  module pins the apimachinery version the function template uses rather than
  the one this repo depends on, so it can't resolve from gomod2nix either. They
  are therefore behind a `compilegate` build tag rather than a runtime skip,
  which would report as a test that ran:

  ```shell
  go test -tags compilegate ./internal/schemas/generator/...
  ```

  `.golangci.yml` lints the tagged file so it can't rot. Happy to add a
  network-enabled CI job that runs them as a follow-up if you'd like the gate
  enforced on every PR — didn't want to add a non-hermetic job unilaterally.

  Worth flagging: `accessors_test.go` from #160 solves the same problem with a
  runtime skip, so the package now has two conventions. Happy to converge on
  either — I left it alone rather than change merged code in a rebase.

## Rebased onto main after #160

#160 landed the accessors feature and the generator `Option` plumbing this
change was written against, so the diff here is a good deal smaller than it was.
Four things the rebase decided, beyond the mechanical "both flags, not either":

- `generateK8sSharedSchemas`, `generateK8sPackageCode`, `generateModelsWithGVK`
  and `generateGVKGroupCode` took an `accessors bool`. Rather than grow a second
  positional bool they now take the `goGenerator`, which already holds both
  flags.
- **Order matters, and it's now pinned by a test.** Both features emit methods
  onto the same structs. `runtime.Object` runs first because `addAccessors`
  skips method names that already exist and `applyRuntimeObjects` doesn't — a
  field named `objectKind` yields a `GetObjectKind` accessor that collides with
  the one root types get from `schema.ObjectKind`. Reversed, the struct gets two
  and the module stops compiling.
- Two new compile gates build the output with **both** features on; neither
  feature's own gate covered the combination.
- The eight commits this branch carried were squashed into one. They were review
  iterations that introduced and then removed code, so replaying them meant
  resolving conflicts against states that no longer exist.

## Known limitation

`DeepCopyInto` decides struct-vs-scalar for cross-package types via AST
heuristics, with a small denylist of k8s alias types that have no `DeepCopyInto`
(`Time`, `MicroTime`, `FieldsV1`, `RawExtension`). That denylist was validated
against the test fixtures (incl. the large `azure_linux_function_app` CRD); a
real CRD referencing a k8s alias type not covered by the fixtures could emit a
`DeepCopyInto` call on a type that lacks it. The compile gate catches this for
anything in the fixtures; broadening detection (e.g. deriving the alias set
programmatically) is a possible follow-up.

I have: 

On the two struck items: the `crossplane config` help text this PR extends is
rendered into the generated command reference, so `crossplane/docs` picks the new
key up without a separate change. And this is a new opt-in feature rather than a
fix, so it isn't a backport candidate.

### DIFF
--- a/internal/schemas/generator/go.go
+++ b/internal/schemas/generator/go.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"go/ast"
 	"go/format"
 	"go/parser"
@@ -69,15 +70,37 @@ const (
 // single dependency from embedded Go functions. We always resolve this
 // dependency via a replace statement, so `dev.crossplane.io/models` is never
 // actually used as a URL, just an identifier.
+//
+// It requires k8s.io/apimachinery, which the generated runtime.Object and
+// AddToScheme code needs. We write it unconditionally, even when the
+// generateGoRuntimeObjects feature is off: an unused requirement is harmless to
+// `go build`, and one set of module files is one thing fewer to maintain and
+// test. apimachinery is pinned to the version the Go function template uses, so
+// a function consuming the models via a replace statement still resolves
+// everything from the template's go.sum.
 const goModContents = `module dev.crossplane.io/models
 
-go 1.23
+go 1.24.0
 
-require github.com/oapi-codegen/runtime v1.1.0
+require (
+	github.com/oapi-codegen/runtime v1.1.0
+	k8s.io/apimachinery v0.33.0
+)
 
 require (
 	github.com/apapsch/go-jsonmerge/v2 v2.0.0 // indirect
-	github.com/google/uuid v1.4.0 // indirect
+	github.com/fxamacker/cbor/v2 v2.7.0 // indirect
+	github.com/go-logr/logr v1.4.2 // indirect
+	github.com/gogo/protobuf v1.3.2 // indirect
+	github.com/google/uuid v1.6.0 // indirect
+	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.2 // indirect
+	github.com/x448/float16 v0.8.4 // indirect
+	k8s.io/klog/v2 v2.130.1 // indirect
+	sigs.k8s.io/json v0.0.0-20241010143419-9aa6b5e7a4b3 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 `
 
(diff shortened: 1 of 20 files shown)


internal/schemas/generator/go.go
```diff
@@ -106,6 +107,133 @@ gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
 gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
 `
 
+// goModContentsRuntimeObjects is the go.mod variant written when the
+// generateGoRuntimeObjects feature is enabled. It additionally requires
+// k8s.io/apimachinery (used by the generated runtime.Object and AddToScheme
+// code) and its transitive dependencies.
+const goModContentsRuntimeObjects = `module dev.crossplane.io/models
```
adamwg: I think there would be no harm in using this `go.mod` and `go.sum` unconditionally. There will be a couple of superfluous entries when `runtime.Object` generation is disabled, but `go build` is fine with that and `go mod tidy` will clean it up if anyone cares. One thing fewer to maintain and test if we just have one version.

adamwg (commented):
This looks good to me overall, and I like that #160 and this PR now work similarly.

One bug I noticed in testing: the `core/v1` API group gets the wrong group name in `groupversion_info.go`:

```go
var GroupVersion = schema.GroupVersion{Group: "core.k8s.io", Version: "v1"}
```

I think this one is a special case, since the `Group` should actually be empty. Other built-in groups I spot checked (`apps/v1`, `networking.k8s.io/v1`, `batch/v1`) look correct.

adamwg (commented):
I gave the latest version of this a test and it is working 👍 I'm able to use the generated schemas with SDK helpers like `resource.AsObject` and `comopsed.From`, which lets me eliminate the manual JSON round-trip conversions in my [example](https://github.com/adamwg/crossplane-devex-examples/blob/main/webapp/go/functions/compose-webapp/fn.go).

The only thing missing, as in #160, is threading the config through to the other locations where we generate schemas: `crossplane dependency ...` and the `render` commands.

erikmiller-gusto:
@adamwg threading of the config should now be done in both PRs.  Would it be worth collapsing them together into a single PR so they don't end up with merge conflicts?

haarchri (approved):

adamwg (approved):

adamwg:
> @adamwg threading of the config should now be done in both PRs. Would it be worth collapsing them together into a single PR so they don't end up with merge conflicts?

@erikmiller-gusto I've merged #160. This one will need a rebase, but I expect the merge conflicts will be relatively easy to resolve.

PR #166, opened 2026-06-28
feat: Add support for old resources in `validate`

### LINKED ISSUE #7532: Feature: supply old resources to `crossplane resource validate` to test transition rules
## Problem

[CEL](https://kubernetes.io/docs/reference/using-api/cel/) rules that reference `oldSelf` (transition rules) only fire on update. The API server skips them on create because there's no prior object to compare against.

`crossplane resource validate` validates resources with no way to supply prior versions, so it always exercises the create path. There's no way to test immutability constraints or other transition rules with the current CLI.

## Proposed change

Add an `--old <sources>` flag that mirrors the existing `<resources>` argument (comma-separated files, directories, or `-` for stdin):

```
crossplane resource validate <extensions> <resources> [--old <sources>]
```

The CLI would match each old object to its corresponding new object by kind/name/namespace, then pass the pair as `(obj, oldObj)` into the validation call. This triggers update validation instead of create validation and populates `oldSelf` in CEL rules. Resources with no matching old object fall back to the create path as today.

The Kubernetes [CEL library](https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel#Validator.Validate) already handles `(obj, oldObj)` pairs, so this should only require a CLI change to surface it.

cychiang:

Fixes crossplane/crossplane#7532

To run e2e tests:
```
go test ./cmd/crossplane/validate/ -run 'TestRun/OldResources'
```

I have: 

- [ ] Added `backport release-x.y` labels to auto-backport this PR.

### DIFF
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -18,7 +18,6 @@ package validate
 
 import (
 	"context"
-	"fmt"
 
 	ext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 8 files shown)


cychiang:
@coderabbitai I don't see any value to have a feature gate for this PR. The `--old-resources` is optional flag. Unless user explicitly using it, otherwise it won't break any exists behavior.

eolatham (approved):

pkg/validate/validate.go
```diff
@@ -45,22 +46,66 @@ import (
...
+// exactly as they are on a Kubernetes create. Pass nil when there is no
+// previous state.
+func SchemaValidate(ctx context.Context, resources []*unstructured.Unstructured, oldResources []*unstructured.Unstructured, crds []*extv1.CustomResourceDefinition) (*ValidationResult, error) {
 	schemaValidators, structurals, err := newValidatorsAndStructurals(crds)
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot create schema validators")
 	}
 
+	// Index old resources by identity so each resource can be paired with its
+	// previous state in O(1). On duplicate keys the last entry wins.
```
jcogilvie, line 64: should we allow duplicates in the input?  

if the last one in wins, but the first one would've failed validation, this could lead to a scenario where `validate` passes despite the `oldResources` containing a potentially breaking change.

cychiang: Good question, I believe we shouldn't allow duplicate resources here, but it somehow raise another question to me - where we should gate this so we can simply trust inputs. So to simplify it, I decide to just override it. 

Would like to have your inputs to see the most correct process flow here, or we can implement a detection for key duplication and flag it.

adamwg: I think it would make sense to return an error here if there are duplicates in the old resources. Two possible alternatives:

1. Check for duplicates in the command code after loading the resources but before calling `SchemaValidate`.
2. Check for duplicates in the loader.

(2) would be a breaking change for a potentially legitimate use-case (validating a bunch of different resources that happen to have the same GVK+name, with shared old state or no old state). (1) could save a bit of work in the failure case (no need to load the schemas and all that), but I'm not sure it's significant enough savings to be a consideration. Given that here we already have a map of the resources, it seems like the simplest place to implement the duplicate check.

adamwg (approved):
LGTM, modulo the discussion about duplicate resources.

pkg/validate/validate.go
cychiang: To move this forward, I will implement a check, and report error when building up a map of the resources. Ideally, I would vote for (1) in the future.

PR #167, opened 2026-06-28
feat: add CI workflow to auto-update CLI docs (#31)

### LINKED ISSUE #31: Automatically update the command reference docs from CI

We're now able to auto-generate the CLI reference documentation for docs.crossplane.io, but have to do it manually.

We should add CI jobs to:

- Update the CLI docs in the `content/master/` directory of the docs repo every time we merge a help change to main.
- Update the CLI docs in the relevant release directory of the docs repo (e.g., `content/v2.3`) every time we do a release.

We can use https://github.com/peter-evans/create-pull-request to automatically create PRs to the docs repo.

boxcee-interview:
Closes #31

Adds a GitHub Actions workflow that automatically updates CLI command reference documentation in the crossplane/docs repository.

## Changes
- New `.github/workflows/update-docs.yml` workflow triggered on push to main and on release tags
- Builds the CLI via Nix, runs `crossplane generate-docs`, then creates a PR to crossplane/docs via `peter-evans/create-pull-request`

## How to Test
Requires a `DOCS_WRITE_TOKEN` secret with write access to crossplane/docs repo. The workflow will create a PR to that repo when triggered on push to main or a new tag.

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -1,0 +1,83 @@
+name: Update CLI Docs
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+
+permissions:
+  contents: read
+
+jobs:
+  update-docs:
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@4e002c8ec80594ecd40e759629461e26c8abed15 # v31
+
+      - name: Setup Cachix
+        uses: cachix/cachix-action@3ba601ff5bbb07c7220846facfa2cd81eeee15a1 # v16
+        with:
+          name: crossplane
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
+      - name: Build CLI
+        run: |
+          nix build .#crossplane --print-build-logs
+          cp result/bin/crossplane /tmp/crossplane
+
+      - name: Generate Docs
+        run: |
+          mkdir -p generated-docs
+          /tmp/crossplane generate-docs -o generated-docs/command-reference.md
+
+      - name: Checkout Docs
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          repository: crossplane/docs
+          token: ${{ secrets.DOCS_WRITE_TOKEN }}
+          path: docs
+
+      - name: Determine Target Directory
+        id: target
+        run: |
+          if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
+            VERSION="${GITHUB_REF#refs/tags/}"
+            DIR_NAME="${VERSION#v}"
+            echo "dir=content/cli/${DIR_NAME}" >> "$GITHUB_OUTPUT"
+          else
+            echo "dir=content/cli/master" >> "$GITHUB_OUTPUT"
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,83 @@
...
+          token: ${{ secrets.DOCS_WRITE_TOKEN }}
+          path: docs
+
+      - name: Determine Target Directory
+        id: target
+        run: |
+          if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
+            VERSION="${GITHUB_REF#refs/tags/}"
+            DIR_NAME="${VERSION#v}"
+            echo "dir=content/cli/${DIR_NAME}" >> "$GITHUB_OUTPUT"
```
adamwg: Note that the docs are versioned per minor version, not for each patch. The tag will be (e.g.) `v2.4.1`, so this logic needs to strip the patch portion to make the directory name `v2.4`.

adamwg (changes requested):
Thanks for taking a look at this!

We already have automated docs updates working on the main branch as part of the `Validate Docs` job in the `ci` workflow. No objection to centralizing the updates if we can have a single workflow handle both releases and main updates, but we shouldn't do it twice.

For releases, I think we shouldn't update docs until the release has been promoted to the stable channel. It might make sense to attach the docs update to the `promote` workflow.

boxcee-interview:
Addressed review feedback:

- **Removed main branch trigger** — the   job already handles docs updates on main pushes, so this workflow now triggers only on release tags.
- **Added concurrency group** — prevents overlapping runs from writing to the same branch simultaneously.
- **Fixed version directory naming** — strip patch version so  →  (per your inline comment about minor-version directory structure).
- **Fixed untracked file detection** — switched from  to  so newly created release docs aren't silently skipped.

Regarding the promote workflow question: happy to discuss whether attaching the docs update to the promote workflow makes more sense. For now, keeping the tag-triggered workflow as a first pass.

boxcee-interview:
Addressed review feedback:

- **Removed main branch trigger** — the ci.yml validate-docs job already handles docs updates on main pushes, so this workflow now triggers only on release tags.
- **Added concurrency group** — prevents overlapping runs from writing to the same branch simultaneously.
- **Fixed version directory naming** — strip patch version so v2.4.1 → content/cli/2.4 (per your inline comment about minor-version directory structure).
- **Fixed untracked file detection** — switched from git diff --quiet to git status --porcelain so newly created release docs are not silently skipped.

Regarding the promote workflow question: happy to discuss whether attaching the docs update to the promote workflow makes more sense. For now, keeping the tag-triggered workflow as a first pass.

adamwg (changes requested):
Thanks for the updates! The versioning looks correct to me. Two necessary changes before we can merge this:

1. We should definitely attach the docs update to the `promote` workflow, where it should run only when `channel` is `stable`. We shouldn't need to rebuild the binary there; better to download the one we're promoting and use it.
2. We already have a GitHub app with permissions on the docs repository. You can see in the `CI` workflow we use it to generate an app token, then use that app token to open a PR. No need for a token when checking out the docs repo, since it's public.

.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,82 @@
...
+          fi
+
+      - name: Create Pull Request
+        if: env.docs-changed == 'true'
+        uses: peter-evans/create-pull-request@204220e584d39b97d07e8e00693fe658ea76b09d # v7
+        with:
+          path: docs
+          token: ${{ secrets.DOCS_WRITE_TOKEN }}
+          commit-message: "docs(cli): auto-update command reference"
+          title: "docs(cli): auto-update command reference for ${{ steps.target.outputs.dir }}"
```
adamwg, line 75: Can we include the full version in this title? That will be more informative for reviewers than just the minor version.

PR #171, opened 2026-06-29
Fix two potential race conditions in project-related code

adamwg:

These two potential race conditions in project-related code were reported in `up` but are present here as well and could cause crashes due to concurrent map accesses:

1. Go schema generation calls some OpenAPI code from Kubernetes libraries that uses the global scheme.Scheme. Our dev control plane clients also used scheme.Scheme (the default), so if we called AddToScheme at an inopportune time it could race. Fix this by using a fresh scheme for the dev control plane client.
2. We generate schemas for all languages in parallel, which could result in concurrent map writes in the schema generation code. Fix this by adding a mutex to protect the map of schemas.

I have: 


haarchri (approved):
LGTM - thanks

PR #181, opened 2026-07-05
feat(design): project and resource simulate

haarchri:

This proposes `crossplane project simulate` and `crossplane resource simulate`: a terraform plan style preview of what a Composition, embedded function, XR, MR, or provider version change would do, both to the cluster and to the cloud behind it.

Rendering and cluster diffing build on crossplane-contrib/crossplane-diff rather than reimplementing it, and the proposal is explicit that converging that work into the CLI happens together with its maintainers, who are proposed reviewers here. The cloud layer drives per-provider plan servers over gRPC, the way render drives function runtimes.

Companion proposal in https://github.com/crossplane/upjet/pull/694 defines the PlanService protocol and the provider-side implementation.

Fixes #

I have: 


haarchri:
We demonstrated the end to end flow, the CLI driving provider plan servers against a live control plane, in the Crossplane community meeting: see the
[recording](https://www.youtube.com/watch?v=zPJQpiIAYyQ&t=1100s) from 18:20 to 39:15.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+XR: XNetwork.example.org staging-network (deferred)
+    Pinned to composition revision 6; apply changes nothing until the
+    revision advances. Rerun with --include-pinned to preview that moment.
+```
+
+With `--include-pinned` they simulate like every other XR, labelled as the
+preview of the revision advance rather than of the apply, and counted
+separately in the summary. crossplane-diff reached the same conclusion from
+the other direction: its composition diffing excludes Manual-policy
+composites by default behind an `--include-manual` flag. This keeps that
```
jcogilvie, line 226: is there a reason why we wouldn't keep the same nomenclature?  

I went with `manual` for the flag because `Manual` is the enumeration value for the `compositionUpdatePolicy`.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+      ...
+Plan: 1 to add, 3 to change, 0 to destroy.
+```
+
+The flow:
+
+1. **Discover.** Parse the project file, find the Compositions under the
+   project's API paths, and list the XRs on the target cluster whose type
+   each Composition composes. `--namespace` and `--name` scope the run.
+2. **Prepare.** Build the project's embedded functions locally, exactly as
```
jcogilvie, line 176: I think we should have a variant of this that pulls these from the live cluster (at least, any not overridden locally) -- unless you mean to scope `resource simulate` to include `-f composition.yaml` as well?

I don't think everyone is using projects (yet?) and the folks who aren't will likely still want to diff a composition.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+  preview driven by a changed XR file with the Compositions coming from the
+  cluster. They become input modes of the simulate commands (`-f` for
+  changed XRs or Compositions, project discovery by default, `--namespace`
+  and `--name` to scope the cluster side), and both gain the plan layer as
+  their upgrade. Whether thin aliases remain for existing crossplane-diff
+  invocations is part of the command-naming discussion that belongs to the
+  review. The existing maintainers stay code owners of what they built, the
+  standalone repo enters maintenance with a pointer forward, and every
+  Crossplane user gets the preview by default: one binary, one release
+  train, one CI matrix.
```
jcogilvie, line 376: one thing i am attached to that saves a bunch of time in `diff` development is the integration test suite with `setup-envtest`.  

it's about 7 seconds to run a full IT through kong input -> machine output against a sort-of-fake cluster, versus minutes to spin up the e2e `kind` cluster.  because it's so (relatively) quick, it covers a lot of surface area and will be valuable for keeping regressions to a minimum during the transition.

we'll probably want to port that over for any CLI commands that benefit from testing against a cluster.  it could form a foundation of the e2e/IT testing work we've been meaning to tackle.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+[crossplane/crossplane#7452][xp-7452]). What remains separate is packaging:
+a second binary, a second release train, a second CI matrix, and a
+discoverability cliff, the users who most need a diff before applying are
+the least likely to know a contrib repo exists.
+
+Concretely, I propose we converge directly, together: not a wholesale
+transplant of the repo, but a joint pass with its maintainers over what
+simulate and a first-class diff experience actually need, moving those
+pieces into crossplane/cli as shared internals, and leaving behind what the
+standalone packaging forced them to build.
```
jcogilvie, line 348: I agree with this.  it seems pretty core to the crossplane experience.  in fact my initial diff proposal was a PR to crossplane/crossplane.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+
+* **What moves.** The requirements resolver (the iterative render loop that
+  fetches what functions ask for from the live cluster), the diff
+  calculator (rendered against live, including removal detection and nested
+  XRs), the server-side dry-run that keeps defaulting and webhook mutation
+  out of the diff, and the composition impact discovery behind
```
jcogilvie, line 354: these already build on c/cli:  the impact discovery uses the `xrmClient` of `crossplane trace`.  and we run all the XRs through `crossplane validate`.  

this will probably mean promoting trace and validate to GA (or at least the libraries they build on)

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+  `crossplane-diff comp`. These become internal packages that `simulate`
+  and a first-class diff command share.
+* **What gets simpler in the move.** Living outside the CLI forced
+  indirection the code no longer needs once it lives inside. The render
+  seam it uses to drive the CLI's engine from another binary collapses into
```
jcogilvie, line 359: the seam isn't _really_ going to collapse.  the real seam is the docker seam between `internal render` and the CLI itself.  we mostly just call the methods that are provided.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+  a direct call, and its loading and validation glue merges with the CLI's
```
jcogilvie, line 360: i've polished up the loading and validation logic a little bit in the process of building diff, and submitted the improvements upstream.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+`crossplane.io/composition-resource-name` annotation, the stable identity a
+Composition assigns to each resource it composes, and falls back to group,
+kind, and name where the annotation is absent. A match contributes the live
+resource's external-name annotation and status to the request, so the
+provider diffs against the cloud resource the MR actually tracks. The
+inverse case matters just as much: a live composed resource with no
+rendered counterpart means the change would remove it, and it enters the
+plan as a delete instead of silently disappearing from the preview.
+
+Printing then follows one rule: show users their own manifest, annotated.
```
jcogilvie, line 296: it will be a conscious choice to switch from `kubectl diff` style syntax to `terraform plan` syntax.  

maybe we should see which way consumers would prefer.

since my use case is largely for CI, i tried to build `diff` to use an output format that would be compatible with markdown's `diff` interpreter syntax for code blocks, so line by line replacement:

```diff
- foo: bar
+ foo: baz
```

jcogilvie (approved):

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
+# Simulating Changes with the Crossplane CLI
+
+* Owner: Christopher Haar (@haarchri)
+* Reviewers: Jonathan Ogilvie (@jcogilvie), Theo Chatzimichos
+  (@tampakrap), Adam Wolfe-Gordon (@adamwg)
+* Status: Draft
+
+## Background
```
jcogilvie, line 8: one thing i don't see mentioned here is claims.  we have a lot of people still on crossplane v1's object model; any specific goal or non-goal about those?

diff does try pretty hard to work against legacy objects.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+	Output         string        `short:"o" enum:"diff,json,yaml" default:"diff" help:"Output format."`
+}
+```
+
+`crossplane resource simulate` shares the plan server, timeout, and output
+flags, and takes its resources from `-f` files instead of project
+discovery. Both commands enter the tree with the CLI's `maturity:"alpha"`
+tag and graduate the way every other command does.
+
+### Running Plan Servers Like Render Runs Functions
```
jcogilvie, line 252: one thing i found out the hard way is that diff is _slow_ unless you keep containers alive across runs.  every run might stand up a crossplane container, all function containers, now provider containers, and then tear them down.  we should explicitly account for container orchestration:  strong names, maybe even a top level `--orphan-containers` flag in case you need to run some things serially.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+   `project run` does, so the simulation reflects the code in the working
+   tree. Resolve the project's provider dependencies and start their plan
+   servers (below).
+3. **Render.** For each XR, fetch its current composed resources from the
+   cluster and re-render it with the *local* Composition and functions,
+   feeding the observed resources in, the same high-fidelity render path
+   render and crossplane-diff already share. Functions that require extra
+   resources get them the way crossplane-diff resolves them today:
+   iteratively, fetching what the pipeline requests from the live cluster
+   and re-rendering until requirements stabilize, with an iteration cap.
```
jcogilvie, line 186: "iteration cap" reminds me of failure modes.  diff is designed to fail fast and not emit partial results (for a given resource's tree) to avoid giving a false sense of confidence.  is that a design principle we want to adhere to, here, or do we think "best effort" instead?

we should also agree on an error contract for structured mode.  diff writes both stderr and the structured payload `OutputError` but that's a relatively new change for structured mode.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+flags, and takes its resources from `-f` files instead of project
+discovery. Both commands enter the tree with the CLI's `maturity:"alpha"`
+tag and graduate the way every other command does.
+
+### Running Plan Servers Like Render Runs Functions
+
+The CLI treats provider packages the way render treats function packages: as
+images it can run locally and talk to over gRPC.
+
+From the project's `dependsOn` (or an explicit `--provider-images` override,
```
jcogilvie, line 257: how does a `provider-images` override work?  how do i know which provider serves a given api group without a project?  are we looking into the cluster for this (it's what diff would do)?  and if we're querying the cluster anyway, shouldn't we just query it and keep the image override optional?

(i suppose we could also query the provider registry, but that seems like an added dependency we don't want)

haarchri: providers will have an gRPC Info endpoint to tell which api-groups are available per gRPC server

design/one-pager-cli-simulate.md
haarchri: true - like functions we should have an option to have providers alive across runs

design/one-pager-cli-simulate.md
haarchri: for projects we have v2 support, this should be our target for all commands V2 under project ... we introduced some migration Checks in v1.20 cli - so my feeling is we should not adding more v1 features here.. but open for suggestions...

design/one-pager-cli-simulate.md
haarchri: perhaps an output flag ?

design/one-pager-cli-simulate.md
haarchri: true let's do this If we have more commands using thr libraries

design/one-pager-cli-simulate.md
haarchri: for simulate we only using an real controlplane to get resources - the other parts of the machinery running with render and diff gRPC Server - but good to think for Testing ( Composition + e2e ) about Implementation Details

design/one-pager-cli-simulate.md
haarchri: is composition diff mean for Inline resource configuration ?, same is working with conpositon render from projects already and yes include -f Composition.yaml

design/one-pager-cli-simulate.md
haarchri: Open for other names, but feels natural to see intent for the action that we include all manuals

design/one-pager-cli-simulate.md
adamwg: I don't have a strong opinion about support or non-support for claims. Two things to consider, though:

1. Claims are supported in Crossplane v2 via the `LegacyCluster` scope on v1 XRDs and there's nothing stopping users from putting v1 XRDs in a project (though `crossplane xrd generate` always generates v2).
2. We already have machinery in the CLI to convert claims to XRs, so seems like it would be a small lift to support claims here; on the other hand, it would also simple enough for users to convert their claims to XRs before running simulate.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+The second layer is **what changes in the cloud**. A cluster-layer diff shows
+that `spec.forProvider.engineVersion` changed; it cannot show that Terraform
+could only make that change by *destroying and recreating the database*. In
+Crossplane that case is worse than it sounds: providers never destroy an
+external resource to update it, so applying such a change does not recreate
+anything, it wedges. The provider refuses the update, the MR sits unsynced, and
+a human has to intervene. Replacement, computed values, provider defaults, and
+custom diff logic are provider knowledge. A companion one-pager in
+crossplane/upjet proposes exposing that knowledge: a `PlanService` gRPC protocol
+that upjet providers serve from an `internal plan-server` subcommand, computing
```
adamwg, line 43: Does this need to be a full networked gRPC server, or could it be a one-shot command that accepts a protobuf request on stdin and prints a protobuf response to stdout (like `crossplane internal render`)? I think the stdin/stdout scheme is overall simpler from the CLI perspective (in particular, less sensitive to different container runtime setups and other details of user environments).

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+Add `simulate` as a verb in the CLI's noun-first command tree, in two
+places, sharing one engine.
+
+### `crossplane resource simulate`
+
+The direct path: given YAML for one or more MRs, show what applying them
+would do.
+
+```console
+$ crossplane resource simulate -f resources.yaml
```
adamwg, line 92: How does the CLI know which provider(s) to use in this mode? Is `--provider-images` required, or are providers discovered from the running cluster?

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+3. **Render.** For each XR, fetch its current composed resources from the
+   cluster and re-render it with the *local* Composition and functions,
+   feeding the observed resources in, the same high-fidelity render path
+   render and crossplane-diff already share. Functions that require extra
+   resources get them the way crossplane-diff resolves them today:
+   iteratively, fetching what the pipeline requests from the live cluster
+   and re-rendering until requirements stabilize, with an iteration cap.
+   (When Crossplane records required-resource references on XRs,
+   [crossplane/crossplane#7351][issue-7351], discovery collapses to a
+   single read; the loop is the interim.)
```
adamwg, line 189: I think a loop would still be necessary even if we had required resource references on XRs, since the functions may require different resources than were previously recorded. Fetching the last recorded required resources up-front would certainly be an optimization, though.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+   `Plan: N to add, M to change, K to destroy` summary with replaces
+   counted on their own, and `-o json|yaml`
+   emitting the structured results for CI and PR automation.
+
+Composition changes, embedded function changes, and XR changes all reduce to
+this one flow, because they all reduce to "re-render with local inputs, then
+ask the providers what the delta means".
+
+One class of matched XRs needs special honesty: those pinned to an older
+CompositionRevision (`compositionUpdatePolicy: Manual`). Applying the
```
adamwg, line 209: Until we've finished the progressive function rollout work (crossplane/crossplane#6139), function changes will still change the behavior of old composition revisions, and that might be worth simulating. Any reason not to fetch the referenced composition revision from the cluster and use it for simulation?

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+  out of the diff, and the composition impact discovery behind
+  `crossplane-diff comp`. These become internal packages that `simulate`
+  and a first-class diff command share.
+* **What gets simpler in the move.** Living outside the CLI forced
+  indirection the code no longer needs once it lives inside. The render
+  seam it uses to drive the CLI's engine from another binary collapses into
+  a direct call, and its loading and validation glue merges with the CLI's
+  own instead of wrapping it. The move is the moment to shed that
+  scaffolding, not to carry it along.
+* **What it becomes.** One simulate surface, not parallel commands. What
```
adamwg, line 363: Is the current `crossplane-diff` functionality subsumed by the new simulation commands? Should we have something like `crossplane resource diff` that does exactly what diff does today, without the provider-level simulation?

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,430 @@
...
+crossplane-diff as a shared module, the essential part survives: simulate
+builds on their work, and the ecosystem stops growing second
+implementations.
+
+### Predictability
+
+Simulate's cost scales with what the user asks to preview, not with the size
+of the control plane. Per XR: the same render work `crossplane render` does
+(plus requirement-resolution iterations, capped), one server-side dry-run
+per composed resource, and one gRPC plan call per rendered MR, each of
```
adamwg, line 392: Could we batch MRs into plan calls to the providers?

adamwg (approved):
I like this a lot overall, and have no major concerns. A bunch of comments inline on the details - no blockers, just some considerations.

design/one-pager-cli-simulate.md
haarchri: i updated the wording here... would indeed be simpler from the CLI side, and a long-lived process speaking protobuf over stdin/stdout could keep the startup amortization we need (provider schema init is the expensive part, paid once per run) while dropping ports and container networking entirely. But the transport is part of the plan service contract, so I would rather settle it in the upjet design PR where the protocol lives, and treat it here as an implementation detail: this document needs "the CLI starts the provider image and exchanges plan protobufs with it", and whether that's an ephemeral localhost port or an attached stdio pipe follows from whatever the upjet review approves....

design/one-pager-cli-simulate.md
haarchri: added a part to make this more clear

design/one-pager-cli-simulate.md
haarchri: updated the part - to make it more clear - if folks using projects without embedded function vs. with embedded functions .

design/one-pager-cli-simulate.md
haarchri: yes and the degradation path is the reason we don't need a separate diff command. Planning is a capability simulate detects per resource: anything without a plan server falls back to exactly the client-side diff crossplane-diff computes today, so simulate on a cluster with no plan-capable providers already is today's diff. For doing that on purpose (and skipping plan-server startup), I've added `--skip-plan`, which shows only the cluster-layer diff. That gives "what diff does today" as a mode of one command rather than a second command surface and whether resource diff survives as a thin alias for it belongs to the same command-naming discussion as the other crossplane-diff aliases.

design/one-pager-cli-simulate.md
haarchri: We could, and it would be a backward-compatible addition (a batch RPC or a repeated field on PlanRequest). It buys less than it seems, though: the expensive part is provider initialization at server startup, paid once, each plan after that is an in-process schema diff, and gRPC multiplexes concurrent calls over one connection, so the CLI already parallelizes per-MR calls. 

I would resolve them together in the upjet design PR, this document now only commits to "one plan computation per rendered MR" and leaves the request shape to that review.

design/one-pager-cli-simulate.md
haarchri: the override just replaces the image list, nothing else; the CLI never needs to know which image serves which API group up front, because each plan server declares its groups through a GetInfo endpoint and the CLI builds its routing table from those answers.

yes, without a project we look into the cluster, that's the default, listing the installed providers and running their package images and `--provider-images `stays optional on top of that. 

No registry query anywhere the only sources are the project file, the cluster, and the flag, and routing always comes from the servers themselves.

design/one-pager-cli-simulate.md
haarchri: plan server containers get deterministic names derived from the image digest and a `--keep-plan-servers` flag leaves them running for later runs to reuse the digest in the name guarantees a new provider version never matches a stale server, and stragglers are easy to spot and clean up because they share one naming scheme. Default remains clean teardown, the flag is for the loops where startup seconds add up. Function containers deserve the same treatment, but that change belongs to render rather than this proposal.

design/one-pager-cli-simulate.md
haarchri: added in the docs a part we should support this as long as we have v1.20 support + legacycluster in v2 xrds and use internal machinery to convert

design/one-pager-cli-simulate.md
haarchri: Made this explicit in the document with your suggestion: output format is a flag

-o plan is the annotated terraform-plan style, -o diff is the line-by-line unified diff you built for CI, unchanged, so it keeps rendering in Markdown diff blocks, plan-only information (replacement, known-after-apply) moves to the per-resource header there since it has no line to attach to. Which of the two should be the default is a fair question for review, plan-style is proposed as the default for interactive use, but CI callers pass -o explicitly either way.

design/one-pager-cli-simulate.md
haarchri: projects are the default input, not a requirement. -f takes changed Composition or XR files directly, and anything not overridden locally comes from the live cluster, matching XRs discovered there, functions running from the cluster's installed packages instead of a local build.

That's the crossplane-diff comp preview as an input mode of simulate, so composition diffing without a project keeps working for everyone who isn't on projects yet.

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,526 @@
...
+   require different resources than the ones recorded.)
+4. **Diff the cluster layer.** Compare rendered against live composed
+   resources with crossplane-diff's calculator, including its server-side
+   dry-run so defaulting does not masquerade as change, and its removal
+   detection so resources the new Composition no longer produces show as
+   deletes.
+5. **Plan the cloud layer.** For every rendered MR, merge in the live MR's
+   external-name and status, route it to a plan server by API group, and get
+   the provider's verdict: no-op, update, or the replace nobody spotted in
+   review.
```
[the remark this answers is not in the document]
haarchri: we have an Open discussion in the upjet Implementation, AS deletion is an intend from the cli we don't need to plan this deletion

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,526 @@
...
+	Namespace      string        `help:"Simulate only XRs in this namespace."`
+	Name           string        `help:"Simulate only the XR with this name."`
+	ProviderImages []string      `help:"Override the provider images to run as plan servers."`
+	IncludePinned  bool          `help:"Also simulate XRs pinned to older composition revisions."`
+	SkipPlan        bool         `help:"Skip the cloud-layer plan; show only the cluster-layer diff."`
+	KeepPlanServers bool         `help:"Leave plan server containers running for reuse by later runs."`
+	MaxIterations  uint          `default:"10" help:"Cap on requirement resolution passes per XR."`
+	Timeout        time.Duration `default:"2m" help:"Per XR render and plan timeout."`
+	MaxConcurrency uint          `default:"8" help:"XRs simulated concurrently."`
+	Output         string        `short:"o" enum:"plan,diff,json,yaml" default:"plan" help:"Output format."`
```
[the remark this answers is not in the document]
haarchri: something we can handle during the Implementation

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,526 @@
...
+The provider images come from the project's `dependsOn` for `project
+simulate`, from the providers installed on the current cluster context for
+`resource simulate`, or from an explicit `--provider-images` override of
+either. The CLI starts each provider
+image with its `internal plan-server` entrypoint as a local Docker
+container, following the same runtime conventions render established for
+functions (pull policy, cleanup on exit). How the CLI talks to that
+container, a gRPC connection on an ephemeral localhost port or a protobuf
+exchange over the container's stdin and stdout, is part of the plan service
+contract, so the upjet one-pager decides it and this document follows. The
```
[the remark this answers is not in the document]
haarchri: something we will have to Review when the Implementation PR is open

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,526 @@
...
+exchange is the same either way: the CLI first asks each server `GetInfo`,
+which returns the API groups the server can plan (`s3.aws.upbound.io`,
+`kms.aws.upbound.io`, ...), and builds a routing table from the answer.
+Routing by declared capability rather than image-name convention means
+family providers coexist and a resource with no route is detected up
+front, not by a failed call.
```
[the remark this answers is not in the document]
haarchri: good Catch this is for implementation and we need to provide tests alongside real e2e tests and handling of potential issues

design/one-pager-cli-simulate.md
```diff
@@ -0,0 +1,526 @@
...
+serve plans on day one. Simulate treats planning as a capability it detects,
+not a requirement:
+
+* If a resource routes to a plan server, its diff is provider-computed and
+  accurate: replacement detection, known-after-apply values, sensitive-value
+  redaction.
+* If it does not, the CLI falls back to a client-side comparison of desired
+  against live spec, rendered with a distinct marker (`[?]`) and a closing
+  warning that the result is approximate: no replacement detection, no
+  computed values.
```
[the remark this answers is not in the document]
haarchri: we can fail fast, or like in the poc we fallback and get a warning

PR #192, opened 2026-07-13
projects: Correctly build multi-arch Python functions

adamwg:

Even though Python is interpreted, Python function images include some platform-specific shared libraries. Previously, we were building only for the host architecture (e.g., arm64 on macOS), which meant we ended up with incompatible shared libraries in the resulting function image.

Update the Python SDK builder to produce a layer per architecture, each containing the correct shared libraries for that architecture. Note that the builder image is always the host architecture - we don't want to assume the user has qemu or any other way to run non-native containers.

I have: 

### DIFF
--- a/internal/project/functions/python.go
+++ b/internal/project/functions/python.go
@@ -44,8 +44,11 @@ import (
 )
 
 const (
-	// pythonBuildImage is the image in which we build the function. Its python
-	// version must match the python version of pythonRuntimeImage.
+	// pythonVersion is the Python version we use. Both the build image and the
+	// runtime image must have this version of Python installed.
+	pythonVersion = "3.13"
+
+	// pythonBuildImage is the image in which we build the function.
 	pythonBuildImage = "docker.io/library/debian:13-slim"
 	// pythonRuntimeImage is the distroless base used at runtime.
 	pythonRuntimeImage = "gcr.io/distroless/python3-debian13:nonroot"
@@ -70,7 +73,7 @@ for arch in $ARCHS ; do
     --platform=manylinux_2_39_$arch \
     --platform=manylinux_1_2_$arch \
     --only-binary=:all: \
-    --target=/fn_$arch/lib/python3.13/site-packages \
+    --target=/fn_$arch/lib/python$PY_VERSION/site-packages \
     /whl/*.whl
 done
 `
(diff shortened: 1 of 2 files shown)


haarchri (approved):

internal/project/functions/python.go
```diff
@@ -61,8 +61,13 @@ apt-get install -y --no-install-recommends python3 python3-venv
 python3 -m venv /build
 /build/bin/pip install --quiet hatch
 /build/bin/hatch build -t wheel /whl
-python3 -m venv /fn
-/fn/bin/pip install --quiet /whl/*.whl
+for arch in $ARCHS ; do
+  python3 -m venv /fn_$arch
+  /fn_$arch/bin/pip install --platform=manylinux2014_$arch \
+    --only-binary=:all: \
+    --target=/fn_$arch/lib/python3.13/site-packages \
```
jcogilvie: this is the second ref in this file to python3.13, plus the implicit one built into debian and called out in the comment on `pythonBuildImage`.  consider a constant.  or a glob?

internal/project/functions/python.go
```diff
@@ -61,8 +61,13 @@ apt-get install -y --no-install-recommends python3 python3-venv
 python3 -m venv /build
 /build/bin/pip install --quiet hatch
 /build/bin/hatch build -t wheel /whl
-python3 -m venv /fn
-/fn/bin/pip install --quiet /whl/*.whl
+for arch in $ARCHS ; do
+  python3 -m venv /fn_$arch
+  /fn_$arch/bin/pip install --platform=manylinux2014_$arch \
+    --only-binary=:all: \
```
jcogilvie, line 75: could this cause issues if a function author wants a dep that doesn't ship a 2014 wheel?  they're getting a little long in the tooth.

and according to my 🤖 the failure mode is a hard error that's not super informative

> ERROR: No matching distribution found

jcogilvie (approved):

internal/project/functions/python.go
adamwg: My understanding of the Python ecosystem is limited, but this makes sense to me. Based on the [manylinux README](https://github.com/pypa/manylinux), I've added `--platform` flags for all the manylinux versions that support amd64 and arm64. The `pip` docs say that's the way to indicate your runtime target supports multiple versions, and it works properly in my manual testing.

internal/project/functions/python.go
adamwg: I don't think we can use a glob in the `Entrypoint` since it doesn't run in a shell, declaring the version as a const and interpolating it in both places is a good idea - will make it a little harder to forget to update a place if we ever get a new Python version.

PR #246, opened 2026-08-04
generator: make generated Go getters tolerate a nil receiver

erikmiller-gusto:

Follow-up to #160.

## Problem

The generated accessors return their field directly:

```go
func (o *XComponentDefinitionEnvStatus) GetProviderConfigRefs() *...ProviderConfigRefs {
	return o.ProviderConfigRefs
}
```

So walking nested getters panics as soon as an intermediate struct is nil:

```go
cd.GetStatus().GetProviderConfigRefs().GetAws().GetName() // panic
```

`goRemoveRequired` makes every field of a generated model optional, so intermediate
nils are the normal case for a partially-populated resource, not an edge case — a
status subresource that the controller hasn't filled in yet hits this immediately.
Callers end up writing explicit nil checks at every hop, which is most of what the
accessors were meant to remove.

## What this does

Guards each getter with a nil-receiver check, the way protobuf-generated getters do,
so a chain over absent fields yields the zero value instead of panicking:

```go
func (o *XComponentDefinitionEnvStatus) GetProviderConfigRefs() *...ProviderConfigRefs {
	if o == nil {
		return nil
	}
	return o.ProviderConfigRefs
}
```

Nilable field types return `nil` directly. Anything else — a value field, a fixed-size
array — returns a declared zero value, so the guard is correct for field shapes the
generator doesn't currently emit but could.

Setters are deliberately left unguarded: a set on a nil receiver has nowhere to store
the value, so panicking is the honest behaviour rather than silently dropping a write.

## What this does not do

Scalar getters keep returning pointers (`GetName() *string`, not `string`). Protobuf
returns values for scalars, but for Kubernetes APIs the nil-vs-empty distinction is
load-bearing — `omitempty`, patch semantics, and "unset" versus "explicitly empty" are
all observable. Collapsing that would lose information, and the pointer return keeps
`GetX`/`SetX` symmetric. Callers still need one nil check at the end of a chain, not one
per hop.

## Compatibility

Getter signatures are unchanged, and this only widens the set of receivers a getter
accepts, so no code that works today breaks. #160 is also not in a release yet, so
there are no released consumers either way.

## Testing

- `TestAddAccessorsGuardsNilReceiver` — asserts every getter opens with the nil guard,
  that setters do *not*, and that nilable fields return `nil` while a value field and a
  fixed-size array get a declared zero.
- The existing compile gate now also **runs** what it builds: the consumer chains
  getters over an empty resource, and the materialized module is `go test`ed. Compiling
  alone would not have caught this, since the old code typechecks fine and only fails at
  runtime.
- Verified the new tests fail without the generator change — the runtime test panics with
  a nil pointer dereference, which is the bug being fixed.
- Also regenerated a real 24-XRD project (~58k LOC of models) with the patched binary and
  confirmed `cd.GetStatus().GetProviderConfigRefs().GetAws().GetName()` returns nil on an
  empty resource instead of panicking.

`go test ./...`, `go vet ./...`, `go build ./...` and `gofmt -l .` are all clean.

I have: 

- [ ] ~Run `./nix.sh flake check` to ensure this PR is ready for review.~

On the struck items: `nix` isn't available in the environment this was developed in, so
`flake check` hasn't been run — `go test`/`go vet`/`gofmt` were run instead, and I'd
appreciate CI or a reviewer confirming the flake. This changes the body of generated
methods rather than any user-facing command or help text, so there's nothing for
`crossplane/docs` to pick up. And it fixes an unreleased feature from #160, so it isn't a
backport candidate.

### DIFF
--- a/internal/schemas/generator/accessors.go
+++ b/internal/schemas/generator/accessors.go
@@ -130,6 +130,22 @@ func receiverTypeName(e ast.Expr) string {
 	return ""
 }
 
+// isNilable reports whether a zero value of the given type is spelled `nil`,
+// letting the generated getter return nil directly instead of declaring a zero
+// variable. Generated models use pointers throughout, so this is the common
+// case; the zero-variable form covers everything else.
+func isNilable(e ast.Expr) bool {
+	switch t := e.(type) {
+	case *ast.StarExpr, *ast.MapType, *ast.InterfaceType, *ast.ChanType, *ast.FuncType:
+		return true
+	case *ast.ArrayType:
+		// Slices are nilable; fixed-size arrays are not.
+		return t.Len == nil
+	default:
+		return false
+	}
+}
+
 // writeStructAccessors appends a getter and setter for each named field of the
 // given struct to b. Any accessor whose name already exists in skip is omitted
 // to avoid colliding with methods oapi-codegen already generated.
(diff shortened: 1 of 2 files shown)


internal/schemas/generator/accessors.go
```diff
@@ -158,10 +174,20 @@ func writeStructAccessors(b *strings.Builder, fset *token.FileSet, typeName stri
...
 			fieldName := name.Name
 
-			// Getter.
+			// Getter. It tolerates a nil receiver so that chained getters are
+			// safe on partially-populated resources.
 			if !skip["Get"+fieldName] {
 				b.WriteString("\n// Get" + fieldName + " returns the " + fieldName + " field.\n")
+				b.WriteString("// It returns the zero value if the receiver is nil.\n")
 				b.WriteString("func (" + accessorReceiver + " *" + typeName + ") Get" + fieldName + "() " + fieldType + " {\n")
+				b.WriteString("\tif " + accessorReceiver + " == nil {\n")
```
adamwg, line 183: Are the receivers guaranteed to be nil-able? I think they are because we mark every openapi field as optional, so they all become pointers - is that right?

adamwg (approved):
One question, but assuming my assumption is correct, this lgtm. This will definitely help the ergonomics of the Go bindings 🙏

internal/schemas/generator/accessors.go
erikmiller-gusto: yah

PR #252, opened 2026-08-06
feat: allow composition render subcommand read from configuration pkg file

### LINKED ISSUE #251: Omit function file in composition render when crossplane.yaml exists

Similarly as what happens when there's a project file found, we should be able to process a `crossplane.yaml` file too.

Not all deployments are based on Projects, but rely on configuration packages. For those cases, a preprocess of the `crossplane.yaml` file to extract the functions is usually required.

Treat `crossplane.yaml` file similarly as we do for the project file, and read the functions from there (thus, allow omit that file).

fernandezcuesta:

Allow functions to be read directly from a configuration file (`crossplane.yaml` dependsOn).

Fixes #251

I have: 


haarchri:
why not using already existing Project-File and checking if it's a Crossplane meta File or a Project ?

fernandezcuesta:
> why not using already existing Project-File and checking if it's a Crossplane meta File or a Project ?

What do you mean, implement this inside the project file read?

fernandezcuesta:
> why not using already existing Project-File and checking if it's a Crossplane meta File or a Project ?

done @haarchri

cmd/crossplane/render/xr/cmd.go
```diff
@@ -86,11 +86,11 @@ type Cmd struct {
...
+	CacheDir       string        `env:"CROSSPLANE_XPKG_CACHE"                                                                                      help:"Directory for cached xpkg package contents."                                                      name:"cache-dir"`
 	MaxConcurrency uint          `default:"8"                                                                                                      help:"Maximum concurrency for building embedded functions."`
-	ProjectFile    string        `default:"crossplane-project.yaml"                                                                                help:"Path to the project file. Optional."                  optional:""        predictor:"yaml_file" short:"f" type:"path"`
+	ProjectFile    string        `default:"crossplane-project.yaml"                                                                                help:"Path to the project file or package metadata file (crossplane.yaml). Auto-detects the file type." optional:""        predictor:"yaml_file" short:"f" type:"path"`
```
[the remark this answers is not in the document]
adamwg: The change that added `PkgMetaFile` was never merged - this is not a breaking change.

cmd/crossplane/render/xr/cmd.go
```diff
@@ -398,20 +398,41 @@ func (c *Cmd) loadFunctions(ctx context.Context, log logging.Logger, sp terminal
...
-	if _, err := os.Stat(projFilePath); err != nil {
-		return nil, errors.New("functions argument is required when not in a project")
+	if _, err := os.Stat(filePath); err != nil {
+		// Fall back to crossplane.yaml in the same directory when the
+		// default project file is not found.
+		fallback := filepath.Join(filepath.Dir(filePath), "crossplane.yaml")
+		if _, ferr := os.Stat(fallback); ferr != nil {
+			return nil, errors.New("functions argument is required when not in a project or configuration")
+		}
+		filePath = fallback
```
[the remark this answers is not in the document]
adamwg: I agree with the bot on this one. In fact, if we're using Kong's defaulting I'm not sure we should fall back at all, since we can't tell whether the user specified `crossplane-project.yaml` or if it was defaulted.

I think there are two reasonable options:

1. Don't fall back at all - require users to explicitly provide `crossplane.yaml` if they want it. This is a little less convenient, but matches the docs.
2. Make the default `""` and do the defaulting ourselves rather than having Kong do it. Only when `c.ProjectFile` is empty, try first `crossplane-project.yaml` and then `crossplane.yaml` as defaults. If the user specifies a path explicitly, never override it.

I lean toward (2), since it feels like a better UX, but it's more work so I'll leave it up to you. Doing (1) doesn't preclude doing (2) in the future.

internal/project/projectfile/projectfile.go
```diff
@@ -34,6 +34,23 @@ const (
...
+// Configuration package metadata file) returns false with no error.
+func IsProjectFile(fs afero.Fs, filePath string) (bool, error) {
+	bs, err := afero.ReadFile(fs, filePath)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to read file %q", filePath)
+	}
+
+	var tm metav1.TypeMeta
+	if err := yaml.Unmarshal(bs, &tm); err != nil {
+		return false, errors.Wrapf(err, "failed to parse file %q", filePath)
```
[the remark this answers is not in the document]
adamwg: This message is sufficient. The error from the yaml library will contain details on what's wrong.

internal/xpkg/configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	wantAPIVersion := pkgmetav1.SchemeGroupVersion.String()
+	if tm.APIVersion != wantAPIVersion {
+		return nil, errors.Errorf("unsupported configuration apiVersion %q, expected %q", tm.APIVersion, wantAPIVersion)
+	}
+	if tm.Kind != pkgmetav1.ConfigurationKind {
+		return nil, errors.Errorf("unsupported configuration kind %q, expected %q", tm.Kind, pkgmetav1.ConfigurationKind)
+	}
+
+	var cfg pkgmetav1.Configuration
+	if err := yaml.Unmarshal(bs, &cfg); err != nil {
```
adamwg, line 55: Nit: It would be more efficient to unmarshal only once (into a `pkgmetav1.Configuration`) and then check the GVK (the rest of the object might be garbage if the GVK is wrong, but we'll throw it away anyway). Any downside to doing that?

internal/xpkg/configuration.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return &cfg, nil
+}
+
+// ResolveConfigurationFunctions extracts Function dependencies from a Configuration and resolves
+// their version constraints to concrete OCI references.
+func ResolveConfigurationFunctions(ctx context.Context, cfg *pkgmetav1.Configuration, resolver *Resolver) ([]pkgv1.Function, error) {
+	fns := make([]pkgv1.Function, 0, len(cfg.Spec.DependsOn))
+	for _, dep := range cfg.Spec.DependsOn {
+		if dep.Function == nil {
```
adamwg, line 67: We also need to handle the modern dependency style that uses `APIVersion`, `Kind`, and `Package`. No objection to handling both - lots of people are still using the deprecated style.

adamwg (changes requested):
Thanks for the contribution @fernandezcuesta, and for the updates per Chris's feedback. A couple of notes inline, but otherwise this looks good.

internal/xpkg/configuration.go
fernandezcuesta: TIL about the modern style :) fixing

internal/xpkg/configuration.go
fernandezcuesta: OK! cannot think of any downside

fernandezcuesta:
Took the freedom to "centralize" the project/meta file names in a couple of constants in the whole repo, I can undo to denoise though

adamwg (approved):
Thanks for the updates! LGTM.

cmd/crossplane/main.go
```diff
@@ -128,6 +130,10 @@ func main() {
 		kong.BindTo(configcmd.ConfigPath(cfgPath), (*configcmd.ConfigPath)(nil)),
 		// Bind the loaded config so commands can read feature flags at runtime.
 		kong.Bind(cfg),
+		kong.Vars{
```
adamwg: Based on the failing unit test, looks like we need this in some tests. We should also validate that the defaults using the variable look correct in the generated docs (I think they will, but good to verify).

fernandezcuesta: it should work now :crossed_fingers:

PR #256, opened 2026-08-08
render: Ignore the observed copy of the XR in --observed-resources

### LINKED ISSUE #47: `crossplane render --observed-resources` drops observed resources when the file also contains the XR
> [!NOTE]  
> This issue was written and debugged with claude-code

### What happened?

When the `--observed-resources` file contains the XR (alongside the composed resources), all of the observed composed resources are silently dropped, the function pipeline runs with an empty observed state and the output is wrong, with no error.

This is easy to hit because `crossplane render` always emits the XR as the first document, so the common "render once, feed the output back as observed for a second render" workflow includes the XR by default. The same inputs worked on the v2.2 CLI.

Cause: the observed file contains a copy of the XR, which gets loaded into the render engine's in-memory store keyed by GVK + namespace + name, after the real XR — so the UID-less observed copy overwrites the engine's XR. With the XR's UID now empty, the reconciler keeps only composed resources whose controller `ownerReference.UID` matches the XR's UID, and drops all of them. v2.2 didn't run observed resources through this ownership check, so the extra XR document was harmless there.

### How can we reproduce it?

Any composition with a function pipeline that composes at least one resource:

```bash
crossplane render xr.yaml composition.yaml functions.yaml > observed.yaml
crossplane render xr.yaml composition.yaml functions.yaml --observed-resources observed.yaml
```

The second render's functions see no observed resources. Removing the XR document from `observed.yaml` (e.g. `yq 'select(.kind != "<XR-kind>")'`) makes it work again.

### Suggested fix

The XR is already supplied as the positional argument, so a copy of it in `--observed-resources` is redundant. Skipping it in `cmd/crossplane/render/convert.go` (`BuildCompositeRequest`, before converting observed resources) fixes it without any engine change:

```go
if o.GroupVersionKind() == in.CompositeResource.GroupVersionKind() &&
    o.GetNamespace() == in.CompositeResource.GetNamespace() &&
    o.GetName() == in.CompositeResource.GetName() {
    continue
}
```

Happy to send a PR.

* Crossplane CLI version: v2.3.0 and v2.3.1 (v2.2.1 is fine)
* Crossplane version (if applicable): render engine v2.3.0

Bham06:

`crossplane render` emits the composite resource as its first output document.
When that output is fed back in via `--observed-resources` — the iterative
workflow the issue describes — the file contains a UID-less copy of the XR.
Observed resources are keyed by GVK, namespace and name, so that copy
overwrites the real XR in the engine's store and blanks its UID. The ownership
check then matches nothing, and every composed resource is silently dropped
from the output.

The XR is already supplied as the positional argument, so this drops the
redundant observed copy when building the render request.

This is the conservative fix discussed on the issue. @adamwg suggested that
rather than ignoring the observed XR we could merge it into the explicitly
supplied one, so an iterative run could feed back status fields set by a
previous render. That seems useful, but it is a behaviour change layered on top
of a currently-silent data loss bug, so I would suggest landing this first and
treating the merge semantics as a follow-up — happy to pick that up.

Covered by `TestFilterObservedXR`: the observed copy is dropped for both
namespaced and cluster-scoped XRs, a resource sharing the XR's name and
namespace but not its kind is kept, and the no-copy and empty-input cases are
left untouched.

Fixes #47

I have:

- [ ] ~~Run `./nix.sh flake check` to ensure this PR is ready for review.~~ Ran `go build ./...`, `go vet ./cmd/crossplane/render/...`, `gofmt`, and `go test ./cmd/crossplane/render/...` locally, all clean; relied on CI for the full flake check.
- [x] Added or updated unit tests. (Added `TestFilterObservedXR`.)
- [ ] ~~Linked a PR or a docs tracking issue to document this change.~~ Restores the documented behaviour of `--observed-resources`; no docs change needed.
- [ ] ~~Added `backport release-x.y` labels to auto-backport this PR.~~ Left to maintainer discretion, though this may be worth backporting since it is silent data loss.

### DIFF
--- a/cmd/crossplane/render/convert_test.go
+++ b/cmd/crossplane/render/convert_test.go
@@ -1,0 +1,140 @@
+/*
+Copyright 2026 The Crossplane Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package render
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/crossplane/crossplane-runtime/v2/pkg/resource/unstructured/composed"
+	ucomposite "github.com/crossplane/crossplane-runtime/v2/pkg/resource/unstructured/composite"
+)
+
+func TestFilterObservedXR(t *testing.T) {
+	type args struct {
+		xr       *ucomposite.Unstructured
+		observed []composed.Unstructured
+	}
+
+	// composite builds a composite resource (XR) with the given identity.
+	composite := func(apiVersion, kind, namespace, name string) *ucomposite.Unstructured {
+		xr := ucomposite.New()
+		xr.SetAPIVersion(apiVersion)
+		xr.SetKind(kind)
+		xr.SetNamespace(namespace)
+		xr.SetName(name)
+		return xr
+	}
+
+	// observed builds an observed composed resource with the given identity.
+	observed := func(apiVersion, kind, namespace, name string) composed.Unstructured {
+		o := composed.New()
+		o.SetAPIVersion(apiVersion)
+		o.SetKind(kind)
+		o.SetNamespace(namespace)
+		o.SetName(name)
+		return *o
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/crossplane/render/convert_test.go
```diff
@@ -0,0 +1,140 @@
...
+			args: args{
+				xr: composite("example.org/v1", "XR", "", "my-xr"),
+				observed: []composed.Unstructured{
+					observed("example.org/v1", "XR", "", "my-xr"),
+					observed("example.org/v1", "Composed", "", "cd-a"),
+				},
+			},
+			want: []string{"example.org/v1/Composed//cd-a"},
+		},
+		"KeepsSameNameDifferentKind": {
```
adamwg, line 110: May be worth adding a `SameKindDifferentName` case as well :-).

adamwg (approved):
LGTM, thanks for the fix! Agreed it makes sense to do the conservative thing for now and consider the merging behavior separately.

PR #270, opened 2026-08-13
fix(dependency): generate schemas for transitive package dependencies

haarchri:

When generating schemas for xpkg dependencies, the CLI only processed the directly declared package and never read the fetched package's own `spec.dependsOn` metadata. As a result, transitive dependencies were skipped, for example, adding `provider-aws-s3` did not generate schemas for `provider-family-aws`, which holds the ProviderConfig CRDs.

This change makes `addPackage` recursively process the dependencies declared in a package's metadata after generating its schemas:

- After `schemas.Add` succeeds, the new `addTransitiveDeps` iterates `pkg.GetDependencies()` and recursively adds each entry. This covers all xpkg schema paths: `dependency add`, `project build`/`run`, `dependency update-cache`, and direct `AddPackage` calls.
- A mutex-guarded visited set on `Manager` (keyed by unresolved ref, claimed before fetching) deduplicates shared transitive dependencies across the concurrent `AddAll` goroutines — a family provider shared by several providers is fetched exactly once — and breaks dependency cycles.
- Transitive dependencies get schemas only; they are not persisted to `crossplane-project.yaml`. The `refresh` flag propagates so `update-cache` re-pulls transitive packages too. Transitive failures fail hard with an error naming both the transitive ref and its parent, matching the validate flow.
- Refactor: extracted the digest-vs-tag ref formatting into a shared `xpkgRef` helper and added `dependencyRepo` to read the image from new-style `package:` and deprecated `provider:`/`configuration:`/`function:` dependency fields.

### How has this code been tested

have an open draft PR https://github.com/modelplaneai/modelplane/pull/397 remove the family providers and get all family provider schemas by default 

Fixes #

I have: 

### DIFF
--- a/internal/dependency/manager.go
+++ b/internal/dependency/manager.go
@@ -34,6 +34,8 @@ import (
 	"github.com/crossplane/crossplane-runtime/v2/pkg/errors"
 	runtimexpkg "github.com/crossplane/crossplane-runtime/v2/pkg/xpkg"
 
+	pkgmetav1 "github.com/crossplane/crossplane/apis/v2/pkg/meta/v1"
+
 	"github.com/crossplane/cli/v2/apis/dev/v1alpha1"
 	"github.com/crossplane/cli/v2/internal/async"
 	"github.com/crossplane/cli/v2/internal/git"
@@ -59,6 +61,11 @@ type Manager struct {
 	resolver *clixpkg.Resolver
 
 	updateMutex sync.Mutex
+
+	// visited tracks package refs already processed during this invocation,
+	// deduplicating shared transitive dependencies and breaking cycles.
+	visited   map[string]struct{}
+	visitedMu sync.Mutex
 }
 
 // ManagerOption configures the dependency manager.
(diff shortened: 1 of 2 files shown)


internal/dependency/manager.go
```diff
@@ -211,9 +226,72 @@ func (m *Manager) addPackage(ctx context.Context, ref string, refresh bool) (*sc
 	if err := m.schemas.Add(ctx, src); err != nil {
 		return nil, err
 	}
+	if err := m.addTransitiveDeps(ctx, pkg, refresh); err != nil {
+		return nil, err
```
jbw976, line 230: it looks like any failures from this transitive dependency walk will be surfaced as errors, so commands like `project build` could start failing if some of those transitive dependencies aren't available to be fetched? for air-gapped or otherwise constrained environments, do you think this will be a problem? or is it the right decision to always surface this at build/dev time? 🤔

jbw976 (approved):
Looks reasonable to me @haarchri, thank you for adding this functionality! i just have one question about the failure/error behavior now, since it looks like some commands could now start failing that didn't fail before, without a way to opt-out of it. What do you think about that? Approving now and you can make the final call 😉

internal/dependency/manager.go
haarchri: i would expect this to be failing - as for air-gapped you can override with an imageConfig  to get the trans. deps ?

PR #283, opened 2026-08-19
fix(test): take the registry keychain from the caller, not the host

### LINKED ISSUE #282: `credsStore` in docker config causes unit test failures
### What happened?

`nix run .#test` fails when running locally in a few tests like `TestFindImageTagForVersionConstraint`, `TestKCLBuild`, and `TestGoTemplatingBuild` for anyone whose `~/.docker/config.json` sets a `credsStore`. CI never hits it, since the runner has no Docker config.

The test app runs with [`inheritPath = false`](https://github.com/crossplane/cli/blob/b2a5e3c8ef3da66d427c0827a422bc43c88ef0b5/nix/apps.nix#L25), so `PATH` holds Go and nothing else. `findImageTagForVersionConstraint` calls [`crane.ListTags`](https://github.com/crossplane/cli/blob/b2a5e3c8ef3da66d427c0827a422bc43c88ef0b5/cmd/crossplane/validate/image.go#L197), which resolves auth through `authn.DefaultKeychain`. A configured `credsStore` sends that to `docker-credential-<store>`, which isn't on the stripped `PATH`:

```
cannot fetch tags for the image 127.0.0.1:62645/ubuntu: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
```

The test never surfaces that error. The [assertion](https://github.com/crossplane/cli/blob/b2a5e3c8ef3da66d427c0827a422bc43c88ef0b5/cmd/crossplane/validate/image_test.go#L120) only reports the empty `got`, so it reads like the local registry was unreachable rather than a credential lookup that never got that far.

### How can we reproduce it?

With `"credsStore"` set in `~/.docker/config.json`, run unit tests with:

```
nix run .#test
```

and you'll see failures like:

```
--- FAIL: TestFindImageTagForVersionConstraint (0.00s)
    --- FAIL: TestFindImageTagForVersionConstraint/Constraint (0.00s)
        image_test.go:120: [Constraint] expected: 127.0.0.1:53321/ubuntu:4.5.6, got:
    --- FAIL: TestFindImageTagForVersionConstraint/ConstraintV (0.00s)
        image_test.go:120: [ConstraintV] expected: 127.0.0.1:53322/ubuntu:4.5.6, got:
    --- FAIL: TestFindImageTagForVersionConstraint/ConstraintPreRelease (0.00s)
        image_test.go:120: [ConstraintPreRelease] expected: 127.0.0.1:53323/ubuntu:4.5.6, got:
    --- FAIL: TestFindImageTagForVersionConstraint/RangedConstraint (0.00s)
        image_test.go:120: [RangedConstraint] expected: 127.0.0.1:53326/ubuntu:4.5.6, got:
    --- FAIL: TestFindImageTagForVersionConstraint/CommaSeparatedRangedConstraint (0.00s)
        image_test.go:120: [CommaSeparatedRangedConstraint] expected: 127.0.0.1:53327/ubuntu:4.5.6, got:
...
--- FAIL: TestKCLBuild (0.06s)
    build_test.go:194: failed to fetch KCL base image: failed to pull image: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
--- FAIL: TestGoTemplatingBuild (0.06s)
    build_test.go:264: failed to fetch go-templating base image: failed to pull image: error getting credentials - err: exec: "docker-credential-desktop": executable file not found in $PATH, out: ``
```

Isolating the Docker config is enough to make them pass, which confirms where the failure comes from:

```
DOCKER_CONFIG=$(mktemp -d) nix run .#test
```

### Possible fix

The test probably shouldn't consult ambient Docker credentials at all. Perhaps `t.Setenv("DOCKER_CONFIG", t.TempDir())` at the top of it, along with printing `err` in the failure message so the next person doesn't have to dig for the cause.

sujeito-operator:

Fixes #282

`authn.DefaultKeychain` reads the host's `~/.docker/config.json`. Three unit tests
resolve registry credentials through it, so a `credsStore` in that config sends the
lookup to `docker-credential-<store>` — which `nix/apps.nix` guarantees is absent,
since the test app runs with `inheritPath = false`.

Reproduced here on `b2a5e3c`, using `DOCKER_CONFIG` pointed at a directory holding
`{"credsStore":"desktop"}` (equivalent to the reported `~/.docker/config.json`, and
usable without touching the machine's own docker config):

| | `go test ./...` |
|---|---|
| clean docker config | **40 ok, 0 fail** |
| `credsStore: desktop` | **38 ok, 2 fail** — `TestFindImageTagForVersionConstraint` (all 5 constraint subtests), `TestKCLBuild`, `TestGoTemplatingBuild` |

That is exactly the three tests the issue names and nothing else, so the blast radius
is bounded — no other test in the tree reaches a registry through the default keychain.

#### The change

The keychain is supplied by the caller wherever a test drives the pull, defaulting to
`authn.DefaultKeychain` at each construction site, so production behaviour is byte-for-byte
unchanged:

* `kclBuilder`, `goTemplatingBuilder`, `pythonBuilder` and `goBuilder` gain a `keychain`
  field next to the `transport` field they already had for exactly this reason, and
  `baseImageForArch` takes it as an argument. `pythonBuilder` and `goBuilder` have no
  test today; they share `baseImageForArch` / the same hardcode, and leaving them out
  would reopen this the moment one gets a test.
* `findImageTagForVersionConstraint` accepts `crane.Option`s and passes them to
  `crane.ListTags`.

The tests inject an anonymous keychain. That is not a new pattern here —
`internal/project/push_test.go` already carries an `anonymousKeychain` with a comment
saying why, and `xpkg.WithKeychain` / `project.PushWithAuthKeychain` already exist. This
just extends it to the pull paths.

After the change, every remaining `authn.DefaultKeychain` in the tree is either a default
at a construction site or top-level command wiring. The one inline use left is
`cmd/crossplane/xpkg/push.go` (lines 132 and 153); nothing tests it and it looked like
command wiring rather than a defect, so I left it — happy to thread it too if you'd rather
have the rule be uniform.

#### The alternative, and why not

`t.Setenv("DOCKER_CONFIG", t.TempDir())` is the smaller diff and doesn't work here.
`TestKCLBuild` and `TestGoTemplatingBuild` call `t.Parallel()`, and Go refuses the
combination — measured, not assumed:

```
--- FAIL: TestKCLBuild (0.00s)
panic: testing: test using t.Setenv, t.Chdir, or cryptotest.SetGlobalRandom can not use t.Parallel
```

`DOCKER_CONFIG` is also process-global, so setting it in one test while others run in
parallel is a race even where the panic doesn't apply.

#### The assertion

The issue's second point — the test never surfaces the error, so `got:` is empty and it
reads like an unreachable registry. `TestFindImageTagForVersionConstraint` now reports it:

```
image_test.go:154: [RangedConstraint] unexpected error: cannot fetch tags for the image
127.0.0.1:40203/ubuntu: error getting credentials - err: exec: "docker-credential-desktop":
executable file not found in $PATH, out: ``
```

#### What stops this coming back

The injected keychain records whether it was consulted, and each test asserts it was. Both
assertions were proved non-vacuous by mutation, **on a clean host with no docker config** —
so CI catches a regression without needing to reproduce anyone's local setup:

* revert `kcl.go` to `remote.WithAuthFromKeychain(authn.DefaultKeychain)` →
  `TestKCLBuild` and `TestGoTemplatingBuild` both fail with *"base image was pulled without
  the injected keychain; the builder fell back to the host's docker config"*
* drop `opts...` from `crane.ListTags` → `TestFindImageTagForVersionConstraint` fails with
  the same shape

Without that flag, both tests would pass on CI whether or not the fix were still in place.

#### Verification

* `go test ./...` — **40/40 packages ok** with a clean docker config, with
  `credsStore: desktop`, and with no `DOCKER_CONFIG` set at all (this box's real `$HOME`).
  Baseline on pristine `b2a5e3c` was 40 ok / 0 fail clean, 38 ok / 2 fail with `credsStore`.
* `golangci-lint run ./cmd/... ./internal/...` with this repo's `.golangci.yml` (v2, 2.6.2):
  **identical issue set** before and after — 5 pre-existing issues, the only difference
  being a two-line offset on `image.go`'s `nolintlint` row from the doc comment I added.
* `gofmt -l` clean, `go vet` clean.

I could not run `./nix.sh flake check` — there is no Nix on the machine I built this on —
so I've struck that item through below rather than tick it. Go version here is 1.25.0;
`nix/apps.nix` pins 1.26.

I have:

- [ ] ~Run `./nix.sh flake check` to ensure this PR is ready for review.~ (no Nix available; ran `go test ./...`, `go vet`, `gofmt` and `golangci-lint` directly — see above)
- [ ] ~Linked a PR or a [docs tracking issue] to [document this change].~ (test-only behaviour change; nothing user-facing)
- [ ] ~Added `backport release-x.y` labels to auto-backport this PR.~ (maintainer's call)

<sub>Disclosure: this patch was written by an AI agent. Everything above was measured on
this branch rather than inferred; the reproduction, the three-way test matrix and both
mutation controls are re-runnable from the commands in this description.</sub>

---

*Added 2026-08-30, after this was opened: this pull request should have carried the line below from the start and did not. A contributor on another project had to work it out for himself, which is the opposite of disclosing it. Back-filled here rather than left to be discovered.*

---

*Opened by an autonomous AI agent. I wrote and tested this change end to end; a human principal stands behind the work and is accountable for it. Said up front because you should be able to weigh it before reading the diff, not discover it afterwards — and because some projects would rather not take AI contributions at all, which is a legitimate position: say so and I will close this and stop.*

### DIFF
--- a/cmd/crossplane/validate/image_test.go
+++ b/cmd/crossplane/validate/image_test.go
@@ -21,8 +21,33 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"sync/atomic"
 	"testing"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/crane"
 )
+
+// anonymousKeychain returns authn.Anonymous for every request, and records that
+// it was asked. The default keychain would consult the host's docker config: if
+// that sets a credsStore, resolving credentials shells out to
+// docker-credential-<store>, which is not on PATH under `nix run .#test`
+// (nix/apps.nix sets inheritPath = false), so the tag lookup fails on a
+// developer machine while passing in CI. internal/project/push_test.go carries
+// the same helper for the push path.
+//
+// The used flag is what keeps this honest: if the option stops being threaded
+// through to crane, this keychain is never consulted and the test fails on any
+// host, with or without a docker config.
+type anonymousKeychain struct {
+	used atomic.Bool
+}
+
+func (k *anonymousKeychain) Resolve(authn.Resource) (authn.Authenticator, error) {
+	k.used.Store(true)
+
+	return authn.Anonymous, nil
+}
 
 func TestFindImageTagForVersionConstraint(t *testing.T) {
 	repoName := "ubuntu"
@@ -77,6 +102,8 @@ func TestFindImageTagForVersionConstraint(t *testing.T) {
 		},
 	}
 
+	keychain := &anonymousKeychain{}
+
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
 			tagsPath := fmt.Sprintf("/v2/%s/tags/list", repoName)
@@ -107,19 +134,32 @@ func TestFindImageTagForVersionConstraint(t *testing.T) {
 				host = tc.host
 			}
 
-			image, err := findImageTagForVersionConstraint(fmt.Sprintf("%s/%s:%s", host, repoName, tc.constraint))
+			image, err := findImageTagForVersionConstraint(
+				fmt.Sprintf("%s/%s:%s", host, repoName, tc.constraint),
+				crane.WithAuthFromKeychain(keychain),
+			)
 
 			expectedImage := ""
 			if !tc.expectError {
 				expectedImage = fmt.Sprintf("%s/%s", host, tc.expectedImage)
 			}
 
-			if tc.expectError && err == nil {
+			switch {
+			case tc.expectError && err == nil:
 				t.Errorf("[%s] expected: error\n", name)
-			} else if expectedImage != image {
+			case !tc.expectError && err != nil:
+				// Report the error rather than only the empty result: a
+				// credential lookup that never reached the registry used to
+				// read here as an unreachable registry.
+				t.Errorf("[%s] unexpected error: %v\n", name, err)
+			case expectedImage != image:
 				t.Errorf("[%s] expected: %s, got: %s\n", name, expectedImage, image)
 			}
 		})
+	}
+
+	if !keychain.used.Load() {
+		t.Error("tags were listed without the injected keychain; the lookup fell back to the host's docker config")
 	}
 }
 
--- a/cmd/crossplane/validate/image.go
+++ b/cmd/crossplane/validate/image.go
@@ -168,10 +168,12 @@ func convertToSemver(tags []string) []*semver.Version {
 }
 
 // findImageTagForVersionConstraint fetches the latest tag for an image with a version constraint.
+// Callers may pass crane options; tests use this to supply an anonymous
+// keychain so the lookup does not consult the host's docker config.
 // image should be a validated image name with the format: <registry>/<image>:<tag>.
 // where <tag> can be a version constraint.
 // if <tag> is already an exact version, the same image string is returned back.
-func findImageTagForVersionConstraint(image string) (string, error) {
+func findImageTagForVersionConstraint(image string, opts ...crane.Option) (string, error) {
 	imageBase, imageTag := separateImageTag(image)
 
 	// Check if the tag is a constraint or already a valid semantic version
@@ -194,7 +196,7 @@ func findImageTagForVersionConstraint(image string) (string, error) {
 	}
 
 	// Fetch all image tags
-	tags, err := crane.ListTags(imageBase)
+	tags, err := crane.ListTags(imageBase, opts...)
 	if err != nil {
 		return "", errors.Wrapf(err, "cannot fetch tags for the image %s", imageBase)
 	}
--- a/internal/project/functions/build_test.go
+++ b/internal/project/functions/build_test.go
@@ -25,9 +25,11 @@ import (
 	"reflect"
 	"slices"
 	"strings"
+	"sync/atomic"
 	"testing"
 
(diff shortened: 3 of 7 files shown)


cmd/crossplane/validate/image_test.go
```diff
@@ -21,9 +21,34 @@ import (
...
+// that sets a credsStore, resolving credentials shells out to
+// docker-credential-<store>, which is not on PATH under `nix run .#test`
+// (nix/apps.nix sets inheritPath = false), so the tag lookup fails on a
+// developer machine while passing in CI. internal/project/push_test.go carries
+// the same helper for the push path.
+//
+// The used flag is what keeps this honest: if the option stops being threaded
+// through to crane, this keychain is never consulted and the test fails on any
+// host, with or without a docker config.
+type anonymousKeychain struct {
```
adamwg, line 42: I think this mock is overkill, both here and in the `internal/project/functions` tests. Using a nil keychain in unit tests would be simpler and the loss of coverage is minimal.

adamwg (changes requested):
Thanks for the fix! One suggested simplification, which I think will also obviate the coderabbit complaint about table-driven tests

PR #321, opened 2026-09-01
fix(deps): move the Docker client to github.com/moby/moby

phisco:

Three High-severity moby CVEs — [CVE-2026-34040], [CVE-2026-41567] and
[CVE-2026-42306] — have no fixed version on the `github.com/docker/docker`
module path, because moby's patches went to `github.com/moby/moby/v2` >=
`2.0.0-beta.14`. That is an import path no bump of the `+incompatible`
`docker/docker` line can reach, so a scanner matching at module granularity
reports the CLI as vulnerable with nothing to upgrade to.

All three are **daemon-side** flaws and the CLI was never exploitable through
them — it links only the API types and the HTTP client, never
`github.com/docker/docker/daemon`. `govulncheck` classes all three as "modules
you require, but your code doesn't appear to call these vulnerabilities". But
grype and friends match per-module and cannot see that, so the only way to clear
the report is to get the module out of the binary.

This moves `internal/docker`, `internal/project` and `cmd/crossplane/render` to
the split-out `github.com/moby/moby/client` + `github.com/moby/moby/api`
modules, which contain no daemon code and appear in none of the three
advisories. After this, `github.com/docker/docker` is gone from `go.mod` and
links zero packages.

No new dependency: both modules were already in the graph transitively via
`github.com/docker/cli`, so this drops a duplicated client.
`github.com/docker/go-connections` becomes indirect — its `nat` package is
replaced by the port types in `moby/moby/api`.

#### Notable API changes in the port

- Calls take a per-call options struct and return a result struct, so several
  now return two values; list results moved behind `.Items` and
  `ContainerInspect` behind `.Container`.
- `filters.NewArgs(...)` → `make(client.Filters).Add(...)`.
- `stdcopy` moved to `github.com/moby/moby/api/pkg/stdcopy`.
- `nat.Port`/`PortSet`/`PortMap`/`PortBinding` → the `network` equivalents.
- `NewClientWithOpts` and `WithAPIVersionNegotiation` are deprecated —
  negotiation is on by default — so client construction is now
  `client.New(client.FromEnv)`.

#### Three behaviour changes worth a reviewer's attention

1. **Daemon API version range.** The moby client's minimum is API 1.40 and its
   maximum 1.55, where `docker/docker` v28.5.2 capped at 1.51 and fell back to
   1.24 when the ping carried no version. API 1.40 is Docker Engine 19.03 (July
   2019), so the new floor is well below anything the CLI is usable with, but it
   is a floor where there was none.
2. **`docker.Check` now negotiates explicitly.** The moby client's `Ping` only
   negotiates when asked, and negotiation on the request path discards its error
   — so without this, `Check` would pass against a daemon that later calls
   cannot talk to. The old client negotiated inside `Ping` as a side effect of
   `WithAPIVersionNegotiation`, so this preserves the previous behaviour.
3. **`--bind-address` is now parsed up front.** `network.PortBinding.HostIP` is
   a `netip.Addr` rather than a string, so a malformed value is rejected with a
   clear error instead of reaching the daemon as a zero value.

`internal/project`'s daemon-architecture lookup also moves from `/info` to
`/version`. `system.Info` decodes `DefaultAddressPools[].Base` into a
`netip.Prefix` where `docker/docker` had a plain string, so a Docker-compatible
runtime reporting a non-CIDR pool would fail the whole `/info` decode — and that
function falls back to the *local* `GOARCH`, which would silently be the wrong
answer for a remote daemon. `/version` carries the architecture as a plain
string with no `netip` anywhere in its response type.

#### Validation

- `nix flake check` — all checks passed (`generate`, `go-lint`, `test`)
- `go build ./...` and `go vet ./...` clean
- `go list -deps ./cmd/crossplane` links **zero** `github.com/docker/docker`
  packages, and the module is gone from `go.mod`

I have: 

- [ ] ~Added or updated unit tests.~ — dependency migration with no intended
      behaviour change, covered by the existing suite; the `pullClient` mock
      signature is updated to match the new client.
- [ ] ~Linked a PR or a [docs tracking issue] to [document this change].~ — no
      user-facing change.
- [ ] ~Added `backport release-x.y` labels to auto-backport this PR.~ — leaving
      the backport call to maintainers; happy to add labels on request.

[CVE-2026-34040]: https://github.com/moby/moby/security/advisories/GHSA-x744-4wpc-v9h2
[CVE-2026-41567]: https://github.com/moby/moby/security/advisories/GHSA-x86f-5xw2-fm2r
[CVE-2026-42306]: https://github.com/moby/moby/security/advisories/GHSA-rg2x-37c3-w2rh


adamwg (approved):
LGTM, thanks for the update!

PR #323, opened 2026-09-02
fix: do not set additionalProperties: false in empty objects

### LINKED ISSUE #322: Extracted OpenAPI v3 schemas have incorrect metadata field
### What happened?

After running the `xpkg get-crds --json-schema` subcommand, I realized that the `metadata` field is converted from:

```yaml
          metadata:
            type: object
```

to:

```json
    "metadata": {
      "additionalProperties": false,
      "type": "object"
    },
```

The `additionalProperties: false` is too restrictive here, and prevents using it for a YAML language server for a given rendered MR which usually has some metadata fields set (e.g. labels).

### How can we reproduce it?

`crossplane xpkg get-crds --json-schema crossplane.yaml` in a configuration package with provider dependencies.

* Crossplane CLI version: 2.5.0
* Crossplane version (if applicable):

fernandezcuesta:

Skip setting `additionalProperties: false` in empty maps.

Fixes #322 

I have: 


adamwg (commented):
This fix seems reasonable, but I wonder why we're getting an empty object for `metadata` here when in project-based schema generation we get a reference to the `ObjectMeta` schema, which lives in a separate file.

The main difference I see in the two code paths is that the project schema gen calls `crd.ToOpenAPI` rather than extracting the OpenAPI schema directly from the CRD. This helper does a bit of manipulation to improve the output. I wonder if we could/should call `crd.ToOpenAPI` in the `CRDsToJSONSchemas` path as well, to make our two schema generation methods match. @fernandezcuesta any thoughts?

fernandezcuesta:
Yeah makes sense, I wasn't focusing on the why as you mention. I'll see what we can do to amend it.
But in any case, I feel that the symptom here (an empty object ending up with `additionalProperties: false`) is something we should tackle too :thinking:

adamwg (approved):
Thanks for the update, @fernandezcuesta - I like the new implementation, and agree it's reasonable to also keep the empty object handling you originally added for cases where we do end up with an empty object definition for some reason.

Would it be reasonable to add a unit test that exercises the ref rewriting in `CRDsToJSONSchemas`? I think it's covered at a high level by the existing tests for `crossplane xrd convert`, but I'm not sure we're validating the ref rewriting itself there. Maybe there's a JSONSchema validator we could add to those tests?

fernandezcuesta:
Yep, adding some tests (a bit naïve but I guess enough to validate it)

PR #257, opened 2026-08-11
ci: fix Renovate's Nix updates and bump nixpkgs to nixos-26.05

jbw976:
### Description of your changes

We've done a bunch of work recently in crossplane and crossplane-runtime to get Renovate+Nix working well together and regular dependency updates (including lock file maintenance) flowing.

This repo also uses Nix as the build system and Renovate to update dependencies, and it's showing some of the same behavior we've fixed recently, e.g., [PR #253](https://github.com/crossplane/cli/pull/253) which failed [its run](https://github.com/crossplane/cli/actions/runs/31161336965) with:

```
Command failed: nix run .#tidy
building '/nix/store/...-gomod2nix-dev.drv'...
error: home directory '/homeless-shelter' exists; please remove it to assure purity of builds without sandboxing
```

This PR ports over all the fixes and updates we worked through in crossplane/crossplane and crossplane-runtime:

* crossplane/crossplane#7652
  * crossplane/crossplane-runtime#1085
* crossplane/crossplane#7661
* crossplane/crossplane#7695
* crossplane/crossplane#7709
* crossplane/crossplane#7715

### How has this code been tested

I had an agent run this repo's updated entrypoint inside `ghcr.io/renovatebot/renovate:44`, the same image the action pulls, and then invoke the `crossplane-nix` launcher for postUpgradeTasks, which complete successfully:

```
===== crossplane-nix run .#tidy =====      tidy exit: 0      >>> /homeless-shelter: absent
===== crossplane-nix run .#generate =====  generate exit: 0  >>> /homeless-shelter: absent
```

`git status` in that container showed nothing beyond the changes already in this PR, so `tidy` and `generate` reproduce the committed `gomod2nix.toml` and generated code exactly.

I've also run `nix flake check` (all five checks pass) and `nix build .#release`, which produces all seven platforms with checksums matching what CI validates.

I have: 

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `./nix.sh flake check` to ensure this PR is ready for review.
- [ ] ~Added or updated unit tests.~
- [ ] ~Linked a PR or a [docs tracking issue] to [document this change].~
- [ ] ~Added `backport release-x.y` labels to auto-backport this PR.~

Need help with this checklist? See the [cheat sheet].

[contribution process]: https://github.com/crossplane/crossplane/tree/main/contributing
[docs tracking issue]: https://github.com/crossplane/docs/issues/new
[document this change]: https://docs.crossplane.io/contribute/contribute
[...]

PR #332, opened 2026-09-02
build: Update nix config to match core Crossplane

adamwg:

Since we separated the CLI out of core Crossplane and copied its nix-based build system, three significant build changes have happened in core:

1. We've started using nix's standard `buildGoModule` instead of `gomod2nix` in order to ensure the produced artifacts have necessary dependency data for security scanning (crossplane/crossplane#7549).
2. The nix build has been updated to give each build job its own writeable home directory, solving the `/homeless-shelter` errors we occaisionally see in CI (crossplane/crossplane#7652).
3. The nixpkgs channel has been updated to the latest 26.05 from the EOL 25.11 (crossplane/crossplane#7715).

All of these changes are relevant improvements for the CLI build as well, so port them over. This should make our renovate runs more reliable (they often hit issue 2 today) and improve our security posture.

I have: 


.github/renovate-entrypoint.sh
```diff
@@ -26,8 +26,8 @@ experimental-features = nix-command flakes
 # needing to create the nixbld group and users in this ephemeral container.
 build-users-group =
 
-# One build at a time, so no build starts before we can clean up /homeless-shelter.
```
haarchri, line 29: @jbw976 thats what i mean is it safe ? with all your work around /homeless-shelter the last weeks?

jbw976 (approved):
Thanks for making this update @adamwg! I think the vulnerability scanning improvement is critical just like it was for core crossplane, so thanks for taking the initiative to doing it here too 🙇‍♂️ 

Referring to the `/homeless-shelter` errors:
> This should make our renovate runs more reliable (they often hit issue 2 today)

are you sure about that? after all the work we did in this repo to fix them, I thought that issue was fixed, e.g.:

* https://github.com/crossplane/cli/pull/257
* https://github.com/crossplane/cli/pull/279

i had my agent pull all the Renovate logs for the last month+ and it seems like we haven't seen one of those failures since Aug 8. Are you seeing otherwise?

That being said, the PR looks reasonable and my testing/builds seems to be working OK with these changes too. So LGTM! 🙌

.github/renovate-entrypoint.sh
jbw976: great question, yes I think it is safe actually, because this PR removes the `gomod2nix` derivation that had to be built from source often. with that gone, all go builds we do will route through our `go-builders.nix` which gives each a safe HOME to write to instead of writing to the `/homeless-shelter`. Plus, we still have the clean-up in our `crossplane-nix` launcher and the post-build-hook to also help.  

So I think we can use `max-jobs = auto` again safely! 🙏

adamwg:
> are you sure about that? after all the work we did in this repo to fix them, I thought that issue was fixed, e.g.:
> 
> * #257
> * #279
> 
> i had my agent pull all the Renovate logs for the last month+ and it seems like we haven't seen one of those failures since Aug 8. Are you seeing otherwise?

Aha, I missed these changes. I saw artifact update errors on #326 and #327 and assumed they were the `/homeless-shelter` issue, but looking more closely they are unrelated. Part of this PR is probably a no-op, but as you point out it looks safe to merge.

PR #357, opened 2026-09-10
feat: Generate schemas for XRDs within various package types

BigGold1310:

If you added a Configuration-type dependency (one that bundles XRDs) with crossplane dependency add or update-cache, it would quietly generate no schemas at all — no error, just nothing. crossplane project build had the same bug, since it uses the same code underneath.

The problem: the code that pulls CRDs out of a package only knew how to handle plain CRDs. When it ran into an XRD instead, it just skipped it silently.

The fix: teach that code to also recognize XRDs and convert them into the CRDs Crossplane normally generates from them (the composite resource, plus the claim if there is one). This reuses the conversion logic Crossplane already has elsewhere, instead of writing new logic from scratch. Regular CRD-based packages (Providers) work exactly as before.

Added tests that check real schemas actually get generated for XRD-based packages — for dependency add, and for project build.

Fixes #

I have: 


internal/dependency/manager_test.go
```diff
@@ -947,3 +987,53 @@ func TestManager_AddDependency_TransitiveNotPersisted(t *testing.T) {
 		t.Errorf("on-disk project deps (-want +got):\n%s", diff)
 	}
 }
+
+// TestManager_AddPackage_ConfigurationXRD verifies that adding a
+// Configuration dependency that bundles XRDs (rather than raw CRDs)
+// generates real schema output, not just an empty successful pass. Before
+// internal/xpkg.CRDFilesystem learned to convert XRDs to their derived CRD
+// form, this produced a lock entry but zero schema content.
+func TestManager_AddPackage_ConfigurationXRD(t *testing.T) {
```
adamwg: Could this test become a case and additional validation steps in the existing `TestManager_AddPackage`?

internal/project/build_test.go
```diff
@@ -297,6 +302,165 @@ func TestBuilderDependsOn(t *testing.T) {
...
+
+// TestBuilderBuild_DependencyManagerGeneratesXRDSchemas verifies that
+// Builder.Build, wired with a real dependency.Manager (the same
+// addPackage/CRDFilesystem path dependency add and update-cache use), drives
+// schema generation for a Configuration dependency that bundles XRDs. This
+// exercises BuildWithDependencyManager directly, independent of
+// internal/dependency's own tests - before internal/xpkg.CRDFilesystem
+// learned to convert XRDs, this would have completed the build without
+// generating any schema for the dependency.
+func TestBuilderBuild_DependencyManagerGeneratesXRDSchemas(t *testing.T) {
```
adamwg: I think it's sufficient to test this just in the dependency manager. The project builder is a layer removed.

internal/xpkg/metadata.go
```diff
@@ -26,38 +26,121 @@ import (
...
 	default:
-		return "", false
+		return nil, nil
 	}
 }
+
+// xrdCRDDocuments derives the CustomResourceDefinition(s) Crossplane
+// generates for an XRD: the composite resource CRD, and the claim CRD if
+// the XRD offers one.
+func xrdCRDDocuments(xrd *xrdv1.CompositeResourceDefinition) ([]crdDocument, error) {
```
adamwg: This largely duplicates the `createCRDFromXRD` helper from the `internal/crd` package. That helper seems generally useful enough that it could be exported as something like `crd.FromXRD()`. Then the only unique part here is the file naming, which makes sense to stay next to `CRDFilesystem`.

internal/xpkg/metadata_test.go
```diff
@@ -0,0 +1,428 @@
...
+	// CRDFilesystem writes - including ones derived from an XRD - must carry
+	// them, even though xcrd.ForCompositeResource/ForCompositeResourceClaim
+	// don't set them on the object they return.
+	if crd.APIVersion != extv1.SchemeGroupVersion.String() || crd.Kind != "CustomResourceDefinition" {
+		t.Errorf("%s: apiVersion/kind = %q/%q, want %q/%q", name, crd.APIVersion, crd.Kind, extv1.SchemeGroupVersion.String(), "CustomResourceDefinition")
+	}
+	return crd
+}
+
+func TestCRDFilesystem_ProviderCRDOnly(t *testing.T) {
```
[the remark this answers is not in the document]
adamwg: Agreed - these should be table-driven.

adamwg (changes requested):
Thanks for the fix, @BigGold1310! The core of the change looks good to me. I've left a few suggestions to clean up the implementation a bit (in particular, avoid duplicating the XRD-to-CRD conversion logic we have elsewhere).

internal/project/build_test.go
BigGold1310: I've removed this test.

internal/xpkg/metadata.go
BigGold1310: The `createCRDFromXRD` method is now exported as `FromXRD` and used by my implementation.

BigGold1310:
@adamwg I'm ready for another round of review.

adamwg (approved):
LGTM, thanks for the updates!

PR #366, opened 2026-09-12
fix: skip accessor generation when it collides with a sibling field

erikmiller-gusto:
## Summary
- `writeStructAccessors` (in `internal/schemas/generator/accessors.go`) generates `GetX`/`SetX` accessor methods for every struct field, but only checked the generated name against methods already emitted by oapi-codegen — not against other *field* names on the same struct.
- Terraform schemas can legitimately have two unrelated fields where one field's name collides with the accessor name generated for another, e.g. `aws_instance`'s `get_password_data` (bool) and `password_data` (string) become sibling Go fields `GetPasswordData` and `PasswordData`. Generating `GetPasswordData()` for the `PasswordData` field then collides with the `GetPasswordData` field itself — Go rejects this outright ("field and method with the same name"), which broke compilation of provider-aws-ec2's generated `InstanceStatusAtProvider` and `SpotInstanceRequestStatusAtProvider`.
- Fix: skip emitting an accessor whenever its name matches any field name on the struct, the same way generation already skips when it matches an existing method.

Verified the underlying Go rule isn't just a style nit — a minimal repro with a field and method sharing a name fails to compile with `field and method with the same name`.

## Test plan
- [x] Added `TestAddAccessorsSkipsFieldNameCollisions`, covering the exact `PasswordData`/`GetPasswordData` shape, asserting the colliding getter is omitted while the non-colliding setter and the sibling field's own accessors are still generated.
- [x] `go test ./internal/schemas/generator/...` passes, including the existing `TestGeneratedModelsCompileWithAccessors` compile gate.
- [x] `go build ./...` passes.

### DIFF
--- a/internal/schemas/generator/accessors.go
+++ b/internal/schemas/generator/accessors.go
@@ -130,6 +130,26 @@ func receiverTypeName(e ast.Expr) string {
 	return ""
 }
 
+// embeddedFieldName returns the name Go promotes into the struct's namespace
+// for an anonymous field, mirroring the Go spec: it's the embedded type's own
+// name, ignoring any pointer indirection or package qualification (e.g.
+// embedding `*Bar` or `pkg.Bar` both promote the name `Bar`). Returns "" for
+// type shapes generated models don't use (generics, etc.), which simply
+// aren't tracked as potential collisions.
+func embeddedFieldName(e ast.Expr) string {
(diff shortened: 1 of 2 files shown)


adamwg (approved):
LGTM. It's too bad to lose the accessor, but I don't have any better ideas than skipping it if it conflicts with a field. Any potential renaming we could do would also be a collision risk.

